PR #11, opened 2020-07-24
Update index.js

febrihidayan:
I just installed Oruga on Nuxtjs.
I found an error `defaults is not defined`  oruga\nuxt\index.js:7:39 and now it's finished.
However, I did not find the nuxt directory. This results in an error `Cannot find module '@oruga-ui/oruga/nuxt'`.
I was forced to do it manually.
Thanks


jtommy:
Ok! Please change options.buefy to options.oruga :)

febrihidayan:
Hmmm, it looks like the error is in `defaults is not defined` 👍

febrihidayan:
> Ok! Please change options.buefy to options.oruga :)

Okey @jtommy  😀

PR #21, opened 2020-08-05
New lerna scripts

### LINKED ISSUE #20: Move scripts in the root `package.json`
I have a proposal. Instead of moving into the core package, we could add some script in the root `package.json`. Someting along:
```
    "test:oruga": "lerna run test --scope oruga",
    "test:all": "lerna run test --stream",
```
This way, it would be possible to test the package directly at the root using `npm run test:oruga`. It could be used to build also.

_Originally posted by @service-paradis in https://github.com/oruga-ui/oruga/pull/17#issuecomment-668714655_

astagi:
Fixes #20

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
       # Install dependencies
       - run:
           name: Install Dependencies
-          command: cd packages/oruga && npm install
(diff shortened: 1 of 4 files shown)


febrihidayan:
Awesome 👏

jtommy:
Awesome !!!

PR #22, opened 2020-08-06
Move `openedTopClass` to `rootClasses()`

coreyworrell:
## Proposed Changes

This allows the input to be styled differently depending on the direction of the dropdown. Ex: input border radius can adjust depending on if the dropdown is showing above or below the input.


coreyworrell:
Note: just duplicating the class on the root element should provide backwards compatibility as well if needed, in case someone is relying on that class being where it currently is.

jtommy:
@coreyworrell Welcome ;) thanks!!

PR #26, opened 2020-08-30
Updates color palette with accessible alternatives that match AA

mateuswetah:
Updates color palette with accessible alternatives that match AA minimum contrast level. Updates some of vuepress classes to match identity.

## Proposed Changes
The new color palette:
**primary**: `#5d8000`;
**danger**: `#ed0030`;
**warning**: `#ad6500`;
**success**: `#008930`;
**info**: `#007aCa`;

I still believe we could have some opinion from other users, but that's my bet on it. I know that the _warning_ yellow is not very nice, though. And also, we may wanna rethink the hero section of the home page, the contrast really didn't work well with the emoji :/

An extra detail, I updated two vuepress classes to match the primary color on the "Show code" section of the documentation.


jtommy:
@elevatebart what do you think about it ?

mateuswetah:
Regarding the yellow, I've checked on other places like Bootstrap and WordPress that they end up using a black color text with the yellow background to solve the contrast issue. Its a bit more of work, thought.

elevatebart:
It makes me think that I need to fix docgen-cli on windows 🤣 
The color changes seem almost inconsequential and will make oruga better. 

I like it.

mateuswetah:
@jtommy and @elevatebart, here is another option: the yellow goes dark text. That would be even better, AAA. And if we want to achieve AAA for the rest of the colors, then we would need to darken things a bit more...

image

This would be the color palette:
**primary**: `#445E00`;
**danger**: `#B60000`;
**warning**: `#f4c300`;
**success**: `#006724`;
**info**: `#005C98`;

To compare with the current proposal:
image

Even thought filled buttons loose a bit of their magic I think this darker colors do a great job for outlined buttons and form inputs. Check this out:
image

elevatebart:
much clearer indeed

mateuswetah:
Went full AAA then guys. Lets see if we get used to it :)

PR #82, opened 2021-02-04
Use input instead of native click for table checkboxes. Fixes #81

### LINKED ISSUE #81: Checkable Table do not store the row when individual row is checked
### Overview of the problem

Just reproduce it here: https://oruga.io/components/Table.html#checkable
Browser:
- Edge: v88.0.705.56
- Chrome: v88.0.4324.104

### Description

When a single row is checked, the row is not pushed to checkedRows so it does not sync it at all (`Total checked` do not change)
However when  bulk rows are checked it works as expected.

### Steps to reproduce

Just reproduce it here: https://oruga.io/components/Table.html#checkable

Capture

(Thx for your amazing work)

astagi:
Fixes #81

### DIFF
--- a/packages/oruga/src/components/table/Table.vue
+++ b/packages/oruga/src/components/table/Table.vue
@@ -170,7 +170,7 @@
                             <o-checkbox
                                 :disabled="!isRowCheckable(row)"
                                 :value="isRowChecked(row)"
-                                @click.native.prevent.stop="checkRow(row, index, $event)"
+                                @input="checkRow(row, index, $event)"
                             />
                         </td>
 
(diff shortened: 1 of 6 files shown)


packages/oruga/src/components/table/Table.vue
```diff
@@ -170,7 +170,7 @@
                             <o-checkbox
                                 :disabled="!isRowCheckable(row)"
                                 :value="isRowChecked(row)"
-                                @click.native.prevent.stop="checkRow(row, index, $event)"
+                                @input="checkRow(row, index, $event)"
```
jtommy, line 173: It's correct but we have to stop propagation otherwise the component will emit row click event

jtommy, line 173: It's correct but we have to stop propagation otherwise the component will emit row click event

astagi: Fixed modifying checkbox and radio propagation

PR #84, opened 2021-02-07
Emit update:modelValue instead of input

### LINKED ISSUE #83: Radio button not compatible with Vue 3 in oruga-next
### Overview of the problem

**Oruga** version: [0.3.2]
**Vuejs** version: [3.0.5]
**OS/Browser**: Arch Linux / Mozilla Firefox

### Description

Radio button is not compatible with Vue 3. It emits an `input` event instead of `update:modelValue`

lsnow99:
Fixes #83 

## Proposed Changes

- Make Radio buttons emit `update:modelValue` instead of `input` event in oruga-next


jtommy:
You are right, my fault during migration from Vue 2 code

PR #94, opened 2021-02-20
Fix typo in fade animation

Youniteus:
This fixes the fade animation (used by the loading component)


jtommy:
@Youniteus thanks!

PR #118, opened 2021-04-02
Migrate this.value to this.modelValue

### LINKED ISSUE #117: UI components using this.value instead of this.modelValue in oruga-next
### Overview of the problem

**Oruga** version: 0.3.5
**Vuejs** version: 3
**OS/Browser**: All

### Description

Some UI components still try to access `this.value` in oruga-next instead of `this.modelValue`. More information on this migration can be found [here](https://v3.vuejs.org/guide/migration/v-model.html).

### Steps to reproduce

Create an `o-slider`, and `o-radio` and use `v-model` to bind a local variable.
Try interacting with the components. 

### Expected behavior

The computed properties should resolve correctly

### Actual behavior

The variable bound to `v-model` will update properly, however, computed properties that reference `this.value` within the component source throw an error, and do not reflect updates visually. (For instance, clicking on a radio button does correctly update the underlying `v-model`, however, the circle does not appear filled in).

lsnow99:
Fixes #117 

## Proposed Changes 

- Migrate `this.value` to `this.modelValue`

### DIFF
--- a/packages/oruga-next/src/components/radio/Radio.vue
+++ b/packages/oruga-next/src/components/radio/Radio.vue
@@ -51,7 +51,7 @@ export default defineComponent({
         rootClasses() {
             return [
                 this.computedClass('rootClass', 'o-radio'),
-                { [this.computedClass('checkedClass', 'o-radio--checked')] : this.value === this.nativeValue },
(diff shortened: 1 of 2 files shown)


jtommy:
Thanks, probably I didn't check all during migration

PR #123, opened 2021-04-14
Update Table.vue to remove Vue warn message

cricketthomas:
This just adds one value to the emits array to stop the vue warning message.

[Vue warn]: Component emitted event "cell-click" but it is neither declared in the emits option nor as an "onCell-click" prop.
image

Fixes #

## Proposed Changes

- Add 'cell-click' value to array 'emits'


jtommy:
Thanks!

PR #151, opened 2021-06-14
Patch skeleton computed Classes

makurtu:
## Proposed Changes

This PR shows that on Skeleton there is no `computed` values so it crashes. I just pushed default classes on `rootClass` and now is working. I don't know if some classes might be added, this code is only a "starting point" on skeleton patch.

### DIFF
--- a/packages/oruga-next/src/components/skeleton/Skeleton.vue
+++ b/packages/oruga-next/src/components/skeleton/Skeleton.vue
@@ -1,8 +1,8 @@
 <script lang="ts">
-import { defineComponent, h } from 'vue'
+import { defineComponent, h } from "vue";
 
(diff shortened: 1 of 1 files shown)


jtommy:
@makurtu the Vue 2 component doesn't need it, what's the error?

makurtu:
@jtommy I didn't tried in Vue2, but seems that needs same thing if it crashes. I just tried with Vue3 (Build source code and try on own project).

jtommy:
@makurtu the problem is in the internal core code and fixed https://github.com/oruga-ui/oruga/commit/a8d4c002f18b01b73d4256cf5c3d3ccf8cbaadc1

PR #165, opened 2021-06-29
Add loading prop to button

macjuul:
Added a loading prop to the button component, useful to convey an asynchronous background task triggered by the button. Similar to the disabled prop, mouse events will be disabled when a button is marked as loading. This prop acts similar to the loading prop in Vuetify.

hK1WHdeD6i

I've not implemented these changes in oruga-next since I believe thats better left up to someone else. If you have any further questions or changes I need to make feel free to let me know :)

## Proposed Changes

- Add a "loading" prop to buttons
  - Provides a loading slot to further customize loader appearence

### DIFF
--- a/packages/oruga/src/components/button/Button.spec.js
+++ b/packages/oruga/src/components/button/Button.spec.js
@@ -87,4 +87,18 @@ describe('OButton', () => {
         })
         expect(wrapper.vm.computedTag).toBe('button')
     })
+
+    it('should set tag to "button" if loading', () => {
+        wrapper.setProps({
+            tag: 'a'
+        })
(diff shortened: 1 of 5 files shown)


jtommy:
@macjuul thanks for the PR.
The idea is correct but it doesn't follow the "guidelines" of the other similar components with loading prop/state, for example table.
We prefer add a loader using the component o-loading and of course a slot as you did but without a spinnerClass etc.

Keep in mind that soon we'll add a way to define a slot in the main Oruga config in order to use a custom component as default slot of a Oruga component.

In other words take a look at Table component and we can discuss how to change the PR 👍

Issue #168, opened 2021-07-08
Vue3  "provide" and / or composable for programmatic usage


cimchd:
### Description

With Vue 2 Options-API programmatic usage of oruga was possible with a global $-instance (this.$oruga). In Vue 3 in combination with Composition-API the provide/inject pattern and / or some composables would be better.

### Why Oruga need this feature

Using the Composition-API a lot of code has no access to the this-context of the component. Therefore the $oruga-instance is not accessible in those situations.

jtommy:
@cimchd can you make an example?
Btw there is in plan to rewrite using composition API but I don't think it's a mandatory way given that Vue 3 support options API too

cimchd:
Sure, here is an example:

**OptionsAPI:**
```vue
<script>
export default {
  mounted() {
    // this context available
    this.$oruga.modal.open(...);
  }
}
</script>
```

**CompositionAPI:**
```vue
<script>
import { onMounted } from 'vue';

export default {
  setup() {
    onMounted(() => {
      // no this context
      this.$oruga.modal.open(...); // fails
    })
  }
}
</script>
```

**CompositionAPI with povide/inject:**
```vue
<script>
import { inject, onMounted } from 'vue';

export default {
  setup() {
    const oruga = inject('oruga');
    onMounted(() => {
      oruga.modal.open(...);
    })
  }
}
</script>
```

**CompositionAPI with composable:**
```vue
<script>
import { onMounted } from 'vue';
import { useProgrammatic } from '@oruga-ui/oruga-next'

export default {
  setup() {
    const oruga = useProgrammatic();
    onMounted(() => {
      oruga.modal.open(...);
    })
  }
}
</script>
```

I found a working solution with a named import of ModalProgrammatic for the modal. But I would prefer one of the solutions above.

jtommy:
Do you want to make a PR about it?

cimchd:
> Do you want to make a PR about it?

Sure, just give me some time... :-)

cimchd:
If this is ok for you, it was just a one-liner. I reused the `$oruga` instance in `app.config.globalProperties` for provide.

jtommy:
Great!!
And what about programmatically components https://github.com/oruga-ui/oruga/blob/f7456e215c413b6b9ab79c8053e53b84f7712157/packages/oruga-next/src/components/notification/index.ts#L56?

cimchd:
> Great!!
> And what about programmatically components
> 
[...]

PR #170, opened 2021-07-10
Feat/provide programmatic usage

cimchd:
Related to  #168

## Proposed Changes (oruga-next only)

- add useProgrammatic composable
- refactor registerComponentProgrammatic

### DIFF
--- a/packages/oruga-next/src/index.ts
+++ b/packages/oruga-next/src/index.ts
@@ -31,6 +31,8 @@ export * from './components'
 export { LoadingProgrammatic } from './components/loading'
 export { ModalProgrammatic } from './components/modal'
 export { Programmatic as ConfigProgrammatic } from './utils/config'
+// export programmatic composable
+export { useProgrammatic } from './utils/useProgrammatic'
 
 // export config plugin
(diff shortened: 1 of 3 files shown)


jtommy:
It looks good but before we have to check that treeshaking works well testing it in a full bundle project and a project with only a couple of components. The final bundle has to be different

astagi:
@cimchd I get this error when I try to build Oruga Next 

```sh
src/index.ts → dist/oruga.min.js...
  1748 |         addProgrammatic(property, component);
  1749 |         // add provide and $oruga (only needed once)
> 1750 |         if (!app._context.provides?.oruga)
       |                                   ^ Unexpected token: punc (.)
  1751 |             app.provide('oruga', oruga);
  1752 |         if (!app.config.globalProperties.$oruga)
  1753 |             app.config.globalProperties.$oruga = oruga;
[!] (plugin terser) SyntaxError: Unexpected token: punc (.)
SyntaxError: Unexpected token: punc (.)
```

cimchd:
I have removed the optional chaining operator and pushed a new version.

PR #171, opened 2021-07-14
deep watch arrays in Table component

pakohan:
When running oruga-next with compat mode in Vue3, it's showing me warnings in the console. See the link they provide along the warnings: https://v3.vuejs.org/guide/migration/watch.html

I had problems with auto-formatting in VSCode, it was changing the whole file. Unfortunately the editorconfig may be missing some settings. But I'm not very confident with the setting up formatters in general. :-/


jtommy:
@pakohan thanks a lot!

PR #186, opened 2021-09-05
Fix the path to load helper in `OTableMobileSort` (oruga-next)

j-a-m-l:
Fixes an error that occurs when using oruga-next alone: it tries to load a helper funcion (`getValueByPath`) from the oruga package.


jtommy:
Thanks a lot

PR #188, opened 2021-09-08
Add 'iconPack' to notification props

TvanSchagen:
fix(notification): add missing 'iconPack' argument to notification props, to allow an icon pack besides the default to be set on this component.


jtommy:
It's already here https://github.com/oruga-ui/oruga/blob/89a411cd6dfd1ced9f9d5a4ac3ae3760d7379e28/packages/oruga-next/src/utils/MessageMixin.ts#L27

PR #199, opened 2021-10-03
Changed deprecated .sync to v-model

andrei0x309:
'.sync' modifier on 'v-bind' directive is deprecated. Use 'v-model:propName' instead  vue/no-deprecated-v-bind-sync

Fixes #

## Proposed Changes

-
-
-


andrei0x309:
Ah, ops, I only now realized that docs are for Vue2, I used to use Buefy for vue2, I only migrated to Oruga for V3 compatibility.

jtommy:
@andrei0x309 yes, the docs works on Vue 2.
Take a look at https://github.com/oruga-ui/oruga-bulma (there are a few open issues to be ready for a first release, any help is welcome 😉 )

PR #230, opened 2021-11-03
docs: Show usage of `iconPack` property to set custom icon pack as de…

rubjo:
## Proposed Changes

- use `iconPack` property in code example instead of `pack` attribute for each `o-icon`, since introduction says "import as default config".


astagi (approved):

astagi:
Thank you @rubjo

PR #234, opened 2021-11-06
fix: rendering of comments in horizontal fields

### LINKED ISSUE #226: Comments in horizontal fields render empty inner fields
### Overview of the problem

**Oruga** version: oruga-next@0.4.6
**Vuejs** version: 3.2.20
**OS/Browser**:

### Description

A html-comment in a horizontal field leads to an additional empty inner field. 
grafik
I think this came with commit 5338da4.
My local fix is a `if (element.type.description === "Comment") {return element;}` in `OFieldBody::render()`. 
That works, but I'm not sure this I a bulletproof fix.

### Steps to reproduce

```
<template>
  <div id="app">
    <o-field horizontal  variant="danger" message="Selected subject is wrong">
      <!-- Label left empty for spacing -->
      <o-button variant="primary">
        Send message
      </o-button>
    </o-field>
  </div>
</template>
```

### Expected behavior

Horizontal field should ignore comments.

### Actual behavior

Horizontal field render a empty field for each comment.

MaTTze:
Fixes #226 

## Proposed Changes
render `VNodes` of type `Comment` or `Text` as they are and do not wrap them in `OField`

-
-
-

### DIFF
--- a/packages/oruga-next/src/components/field/FieldBody.vue
+++ b/packages/oruga-next/src/components/field/FieldBody.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { defineComponent, h, resolveComponent } from "vue"
+import {defineComponent, h, resolveComponent, Comment, Text} from "vue"
 
 export default defineComponent({
     name: 'OFieldBody',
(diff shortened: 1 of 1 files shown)


jtommy:
Good catch!

PR #235, opened 2021-11-08
fix: missing 'active-change' event in Datepicker emits

jacoborus:
This change prevents the warnings when opening/closing the datepicker


jtommy:
Thanks

PR #236, opened 2021-11-09
feat: support ligature icon fonts

MaTTze:
## Proposed Changes

- support icon packs based on font ligatures (i.e. [Material Icons](https://fonts.google.com/icons))


jtommy:
@MaTTze thanks but at the moment we don't support all icon pack indeed you can use `iconComponent`

jtommy:
@MaTTze I close it but we can think a way to support better icon packs or a new way to do it

PR #238, opened 2021-11-11
fix: add missing parent param to programmatic loading component

Tomaszal:
This change makes programmatic loading component consistent with modal and notification components. It is needed as correct parent context may be necessary if a custom icon component is defined.


jtommy:
@Tomaszal thanks

PR #242, opened 2021-11-14
fix: set tabbed transitions correctly

### LINKED ISSUE #241: Steps: wrong transition direction
### Overview of the problem

**Oruga** version: all
**Vuejs** version: all
**OS/Browser**:

### Description

The step content always transitions from right to left ever when clicking "prev".

### Steps to reproduce

Navigate through the steps in the Steps-Documentation/example 

### Expected behavior

The step content transitions from right to left ever when clicking "next" and from left to right when clicking "next".

### Actual behavior

The step content always transitions from right to left ever when clicking "prev".

MaTTze:
Fixes #241 

## fixed problems

- `oldTab` in `performAction()` always was the `activeItem`, so the index was always the same in `deactivate()`/`activate()` and therefor transistion name always `slide-prev`
- `performAction()` was always called 4 times (2x `childClick()`, 2x `modelValue()`-watcher

### DIFF
--- a/packages/oruga-next/src/utils/TabbedMixin.ts
+++ b/packages/oruga-next/src/utils/TabbedMixin.ts
@@ -65,9 +65,9 @@ export default (cmp: string) => defineComponent({
          * When v-model is changed set the new active tab.
          */
         modelValue(value) {
-            this.performAction()
-            this.activeId = value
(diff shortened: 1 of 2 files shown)


jtommy:
Thanks @MaTTze

PR #253, opened 2021-12-13
fix inputitems dropdown width

### LINKED ISSUE #252: inputitems' dropdown gets smaller as items are added
### Overview of the problem

**Oruga** version: 0.4.8
**Vuejs** version: 2.6.14
**OS/Browser**: WIN11/chrome  96.0.4664.93

### Description

the input's size gets narrower as more tags are added, which is [synced to the dropdown's width after every click](https://github.com/oruga-ui/oruga/blob/5e02268c7f9c3972a857063f14f28bb0cfbad762/packages/oruga/src/components/autocomplete/Autocomplete.vue#L774)

### Steps to reproduce

1. start selecting items at [inputitems / autocomplete](https://oruga.io/components/InputItems.html#autocomplete)
2. notice dropdown width keeps shrinking

### Expected behavior

dropdown width stays constant regardless of tags selected

### Actual behavior

dropdown width keeps shrinking

laszlo-kasa:
Fixes #252

## Proposed Changes

- added an optional internal/undocumented `HTMLElement` prop to `Autocomplete` to match the dropdown width with instead of the input which kept shrinking after adding tags


laszlo-kasa:
> The fix is great but I don't like the approach, maybe it's better always use container as full width

I don't get it, could you please elaborate?

jtommy:
I added it due to `append-to-body` prop, usually it doesn't need a calculated with

laszlo-kasa:
> I added it due to `append-to-body` prop, usually it doesn't need a calculated with

i see, but this is also broken with `append-to-body` i suspect

jtommy:
Yes, I have to rewrite that part.
What do you think to disable style binding when is not append to body?

PR #258, opened 2021-12-26
Accessible tabs

lsnow99:
## Proposed Changes

- Allow the tabs component to be navigated via left/right/up/down arrow keys
- Add the tab and tablist roles for screen readers

Guidelines here: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html


jtommy:
@lsnow99 I'm not so expert but it works well, thanks!

PR #273, opened 2022-01-18
build: use `.mjs` extension for built esm files

danielroe:
This PR (which will be a breaking change) changes the file extension for esm builds to `.mjs` to enable native node esm imports (for frameworks like nuxt 3).


jtommy:
@danielroe thanks a lot 👍

PR #286, opened 2022-02-05
fix: allow access to full button width

garymathews:
Content inside a button is regarded as basic text meaning we cant make use of the full button space. For example we can't use styling to have left and right content as `o-btn__wrapper` and `o-btn__label` prevents our slot accessing the full button width.

I created an example to demonstrate this by overriding those classes.

##### Example
https://codesandbox.io/s/oruga-sandbox-qv5gt?file=/src/components/HelloWorld.vue

### DIFF
--- a/packages/oruga-next/src/components/button/Button.vue
+++ b/packages/oruga-next/src/components/button/Button.vue
@@ -15,11 +15,8 @@
                 :both="iconBoth"
                 :class="iconLeftClasses"
             />
-            <span
-                :class="labelClasses"
-                v-if="label || $slots.default">
-                <slot>{{ label }}</slot>
-            </span>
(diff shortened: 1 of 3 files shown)


jtommy:
I understand your need but it looks like a breaking change.
Probably not for users that are using full-theme but in general it allows to style the content in different ways due to slot or label prop.

garymathews:
From what I can see, the only breaking change is that the `labelClass` would not be applied to the slot. Which can be addressed.

jtommy:
Yes, I meant it

jtommy:
Yes! Your PR will "break" all apps that are using a button with slot and `labelClass`

jtommy:
@garymathews not it looks ok. It's no a breaking change and it should be ok for your need

PR #289, opened 2022-02-08
New Feature: Datepicker: Let user override < > navigation by month or year

Max1us:
## Proposed Changes

On the datepicker component, introduce a new optional prop that would let the user override < > navigation by month or year
-
-
-


jtommy:
You can customize it by `header` slot in the datepicker instance or creating a own datepicker that extends and use oruga datepicker. 

Alternatively you might add two buttons (default not visible) to switch year, for example:

image

PR #300, opened 2022-02-16
[rfc] chore(dropdown): Add a property to configure dropdown/autocomplete menu tag

elthariel:
## Proposed Changes

- Add a `menuTag` property to the Dropdown component. It'll help me use Oruga with DaisyUI which require the dropdown to be a li
- Add a `menuTag` and `itemTag` to the Autocomplete component, for the very same reason

I need some guidante to write tests on this

### DIFF
--- a/packages/oruga-next/src/components/autocomplete/Autocomplete.spec.js
+++ b/packages/oruga-next/src/components/autocomplete/Autocomplete.spec.js
@@ -213,4 +213,10 @@ describe('OAutocomplete', () => {
 
         expect(subject).toBeFalsy()
     })
+
+    it('has configurable menu and item tags', () => {
+        wrapper.setProps({menuTag: 'ul', itemTag: 'li'})
+        wrapper.find('ul.o-acp__menu').exists()
+        wrapper.find('li.o-acp__item').exists()
+    })
 })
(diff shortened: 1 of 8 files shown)


elthariel:
I'm actually not sure it solves my problem. I'll confirm when I manage to get this modified package in my project :sob:

elthariel:
Yup, it does the job as expected <3

elthariel:
It seems the documentation isn't including the new property. Is there something that I missed ?

jtommy:
> It seems the documentation isn't including the new property. Is there something that I missed ?

You have to run the docs to build it.
Also you are working only on Vue 2 component, you should edit oruga-next package.

elthariel:
@jtommy I was hoping they were somehow auto-magically generated. I'll do that right away, alongside some similar customization in the autocomplete component

elthariel:
I've added some specs \o/. Didn't realized jest was so close to ruby's rspec.

jtommy:
I understand your need but if we meet all CSS framework requirements the library is not si generic.

elthariel:
I totally get your point, it's up to you. 

The only point I could vaguely add on that case is that it allows to add some level of semantic information to the page. The dropdown menu and the autocompletes are inherently list items and it makes sense to be able to express them as such. But it's sort of a pretty weak hand-wavy point :smile_cat: 

I won't feel bad if you decide it's not relevant enough, I'm happy enough with the process of contributing something in a new stack and wanted to give something back, even if insignificant to this great work I'm having a lot of fun with. <3

elthariel:
<3

jtommy:
@elthariel please check your email, I have a proposal for you 😃

Issue #225, opened 2021-10-28
Add Dialog Component


mouadTaoussi:
### Description
Adding Dialog Component .

### Why Oruga need this feature
Easiest way to propmt or alert the user

**Note:I Might take time becuase I have a Display issue in my Laptop that needs to be fixed first, then work on the feature!**

Thaurin:
Yes, I need this before I can finish migrating to Oruga!

Thaurin:
This is literally keeping me from migrating a Vue 2 web application that is in production to Vue 3 with Oruga with Bulma plugin, then Vue 3!

jtommy:
I got it, in the next release will be available the carousel.
There is a PR but incomplete and I need time to add Oruga features

urkle:
@jtommy @mouadTaoussi so I took a look some at the PR #254 and it currently doesn't really work.  
I have a local branch where I have done some rework on it and have it doing the basic alert dialog now (including working with theme-bulma, but only Vue2 right now)

@jtommy A few questions I have though on how you envision this component working in Oruga.

1) the dialog currently uses them-bulma-specific inner contents (modal-card-head, modal-card, etc..).
   Q: how do you envision exposing this to configuration to allow the best customization by the CSS framework layer?
2) the dialog currently "extends" off of Modal and re-implements a lot of properties and code.
   Q: right now my version just has an o-modal in the body that I am using.  Is that acceptable? or would you still rather have it extend OModal?
What is the "official" set of properties it should expose to the $dialog.* methods?

jtommy:
1. As i told you in this PR there is nothing about Oruga idea. For example there isn't bodyClass prop or no modal class are applied to it. Does it make sense?

About template I don't know.. dialog might be different by modal but usually is only a built-in modal with a template for alert, confirm and prompt.

urkle:
I understand that PR was about a quick and dirty port..  I'm trying to make a PR that is more close to the final "oruga style" idea.  Hence all the questions..  I'd rather build up something useful that can easily be included in the next release vs something that'd have to be redone again to fit more with the Oruga style.

As for why add a dialog component?  Because there is some nicety in having a built-in alert, confirm, and prompt style dialogs in the system.  AND especially if those work in a promise/async+await style manner to make coding easy.  

Or find a way to "easily" provide that kind of thing using the programatic modal? and include some nice examples on how to utilize that? (e.g. like the "tost" examples for notifications)


In the end I am wanting something like this.

```javascript
[...]

PR #314, opened 2022-03-13
add a promise return to the programmatic modal open.

urkle:
Proposal for #225

Depends on PR #315 

## Proposed Changes

- Modal instance now has a "promise" property that when using $oruga.modal.open contains a promise that will resolve to the value passed to the $emit('close', value)
- when a modal is canceled, the promise resolves {action: 'cancel', method: 'x'}

### DIFF
--- a/packages/docs/components/Modal.md
+++ b/packages/docs/components/Modal.md
@@ -143,9 +143,146 @@ title: Modal
       },
       cardModal() {
         this.$oruga.modal.open({
+          // parent is only for Vue2. in Vue 3 omit this option
           parent: this,
           component: ModalForm,
           trapFocus: true
+        })
+      }
+    }
+  }
+</script>
+
+<style>
+  .modal-card {
+    display: flex;
+    flex-direction: column;
+    max-height: calc(100vh - 40px);
+    overflow: hidden;
+  }
+  @media screen and (min-width: 769px) {
+    .modal-card {
+      margin: 0 auto;
+      max-height: calc(100vh - 40px);
+      width: 640px;
+    }
+  }
+  .modal-card {
+    margin: 0 20px;
+    max-height: calc(100vh - 160px);
+    overflow: auto;
+    position: relative;
+    width: 100%;
+  }
+  .modal-card-foot,
+  .modal-card-head {
+    align-items: center;
+    background-color: #f5f5f5;
+    display: flex;
+    flex-shrink: 0;
+    justify-content: flex-start;
+    padding: 20px;
+    position: relative;
+  }
+  .modal-card-head {
+    border-bottom: 1px solid #dbdbdb;
+    border-top-left-radius: 6px;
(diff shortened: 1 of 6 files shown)


jtommy:
The idea is good but it's a breaking change.
I developed a similar feature in Buefy dialog in order to support Promise without introducing a breaking change, try to take a look :)

jtommy:
It's always a breaking change because Promise is not supported by all browsers (Vue 2 supports IE11).
We might think to introduce it in the next breaking only for next version, I think Buefy dialog implementation is a good compromesso, don't you?

urkle:
> 

Since it only extends the Modal (e.g. it doesn't change the return contract) we can do one of the following
1) have an option to turn it on an off (e.g. whether I set the promise property or not)
   - but I don't like changing the return contract based on a global variable lik the Buefy implementation does as that means you either can have a promise OR gain access to the instance but never both.
2) if there is no "Promise" available then we don't set it.

IMHO option 2 sounds the best.  As in reality anyone who  is using "promises" in their code (or async/await) and wanting to support IE 11 will have a polyfill in their code builds so things will just work.

I can do a quick change for that.

urkle:
@jtommy I added a guard around the promise usage for the Vue 2 implementation (not the Vue 3).

Further I added extra comments in the example to clarify that promises can't be used on IE 11 unless a polyfill is added to your build.

I also added a comment in the example to show that "parent: this" is only needed for Vue 2 and should not be included in Vue 3.

jtommy:
@urkle I agree with you.
I am not sure about the change of programmatic prop but it works.
After that PR, are you able to implement a dialog? For example a prompt?

urkle:
>  I am not sure about the change of programmatic prop but it works.
Since it wasn't even documented, I figured it was a simple way to pass those callbacks into the dialog easily. If you prefer I can add it as a different prop name, but utilizing the programmatic was simple (though I need to update the property definittion for the IE11 tweak)

> After that PR, are you able to implement a dialog? For example a prompt?

Yes, with this change I can implement my dialog prompts the way I want.  Actually I'm going to "amend" this PR once more before you merge it to separate out the examples so that there is a separate "programmatic" example using promises from the one that is basic. And I'll have that later one implement a simple "prompt" dialog.

Now, a followup question,
Can we apply this "pattern" to notifications and loading? (so developers can essentially tab into a async/await approach to handling when it is closed/dimissed).

urkle:
@jtommy all pushed up.

Changes I made since the last one are
1) adjusted "type" of promise property so that Vue doesn't yell about it (it's undefined meaning any for now. we *could* use Promise in the vue3 one if we wanted)
2) I split the programmatic promise/async/await example to a separate example grouping and restored the previous one. The new example shows a basic implementation of a "prompt" (yes/no) style dialog.  It will "return" either action: 'yes', action: 'no', or action: 'cancel' (if they hit off the dialog or hit the X icon).

PR #328, opened 2022-03-24
Feature/close all notifications

### LINKED ISSUE #322: Close all shown toasts/notifications
### Description

Ability to close all shown toasts/notifications.

Useful when route changes. Clear page from all shown toasts/notifications when route changed.

urkle:
Fixes #322 

## Proposed Changes

- adds a $oruga.notifications.closeAll() method that will close all programmatic notifications

### DIFF
--- a/packages/docs/components/Loading.md
+++ b/packages/docs/components/Loading.md
@@ -155,9 +155,14 @@ title: Loading
 <template>
   <div>
     <o-field>
-      <o-button size="medium" variant="primary" @click="openLoading">
-        Launch loading
-      </o-button>
+      <div class="buttons">
+        <o-button size="medium" variant="primary" @click="openLoading">
+          Launch loading
+        </o-button>
+        <o-button size="medium" variant="primary" @click="openLoadingCloseAll">
+          Launch loading (closeAll)
+        </o-button>
+      </div>
     </o-field>
     <o-field>
       <o-switch v-model="isFullPage">Display loader over full page</o-switch>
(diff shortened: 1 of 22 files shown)


urkle:
@jtommy should we also extend this concept to programmatic modals and loading components?

jtommy:
> @jtommy should we also extend this concept to programmatic modals and loading components?

Yes

I'm reviewing the PR, programmatic component looks like a container of any thing 😄 
Btw why don't you add the instance into index.ts/.is in the open method?

urkle:
> > should we also extend this concept to programmatic modals and loading components?
> 
> Yes
OK.. I'll build a followup PR to this one that adds it to those as well.

> I'm reviewing the PR, programmatic component looks like a container of any thing 😄 
Yea, basically.  If all we needed was a "truthy" to determine if we are programmatic, why add more properties for setting the extra data needed for extra functionality. Just change programmatic to be an object :)

> Btw why don't you add the instance into index.ts/.is in the open method?
So, yeah I could have done that in there.. but it felt "wrong" to have index add it and the component remove it.  It felt better to let the component add and remove itself instead so it was the only one responsible.

jtommy:
@urkle does it ready to review?

jtommy:
@urkle did you try it using individual imports? For example importing only notification and modal and not all Oruga bundle?

urkle:
@jtommy I just did that right now importing just the pieces and injecting them like so
```javascript
  createApp(App)
    .use(Modal)
    .use(Button)
    .use(Notification)
    .use(Loading)
    .use(Config, bulmaConfig) // load theme
    .mount('#app');
```

Also calling closeAll on the imported NotificationProgrammatic module works as well.

PR #349, opened 2022-04-24
correct promise property types

urkle:
## Proposed Changes
- this silences Vue warnings about the wrong data type passed to the property
- for Vue-2 oruga we have a hack that sets Promise to window.Promise so we can accommodate IE 11 which may not have a promise (unless the developer polyfills it)


jtommy:
For Vue 2, probably it's better to create a custom  function that returns a Promise (if the browser supports it) or a simple object. 
About docs don't worry, the `@ignore` doesn't add the prop to the docs 😉

urkle:
> For Vue 2, probably it's better to create a custom function that returns a Promise (if the browser supports it) or a simple object. 

I can adjust that for Vue2.

> About docs don't worry, the `@ignore` doesn't add the prop to the docs 😉 (run docs gen you can see that prop won't be added)

I don't think that is working then.   as I still see the props when I generate docs (even the public docs on oruga.io show the props)

urkle:
@jtommy I reworked to use a helper function.  and I also manually removed the programmatic and promise from the generated docs from `vue-docgen` as it is not respecting the @ignore like their docs claim it is supposed to. Now when I run vuepress build it correctly does not have those two props documented.  However, there are other ignored props that are included in their respective docs still :( (e.g. iconBoth from the Button component)

urkle:
@jtommy I pushed another commit to this PR that fixes the doc generation so that it properly ignores ignored props.   The issue seemed to be around your customization of the prop template. I added a filter loop in there that removes the ignored items.

PR #365, opened 2022-05-26
fixed o-field: message slot not passed down on o-field-body

amicus-veritatis:
Fixed o-field: message slot not passed down if horizonal === true

## Proposed Changes
- FieldBody.vue checks this.parent.hasMessageSlot
- if (this.parent.hasMessageSlot AND first) => render message slot


jtommy:
It looks like a breaking change and also it removes `p` using a slot.
What happens in a form with a mix of message slot and message prop?

amicus-veritatis:
1. I intended to remove ```p``` wrapper from message slot for following reasons:
   - Accessibillity concern: Current ```p``` wrapper lacks several wai-aria attributes, such as ```role="alert"```, ```aria-label``` or ```aria-labelledby``` or etc. 
   - There could be a case where an user want to insert their message slot into flex wrapper; like tooltip.
   - In general, it would make customization much easier without headache. 
 2. Currently, ```messageClass``` doesn't pass down to ```p``` tag if ```horizontal===true```. While ideally this could be solved by pass down ```this.parent.messageClasses``` to FieldBody.vue, I considered it also can be solved by using message slot. 
 3. In a form with a mix of message slot and message prop, messageSlot takes precedence. but it does not prevent user from inserting message into message slot itself; like following code. However, I did not rule out message props completely since it can be accessed via Javascript. 
 ```js
 <o-field horizontal label="Age" name="age" id="fieldAge" message="This message will be ignored">
 <template #message>
 <a href="https://examplewebsite.exampletld" title="Error occured">
 <p role="alert" :aria-label="customMessage">
  {{customMessage}}
 </p>
 </a>
 </template>
 <input type="text" inputmode="numeric" pattern="[0-9]*">
</o-field>
 ```

jtommy:
I understand your point but keep in mind that a user that uses message prop in a field and message slot in on other field could be see a different style (margin, padding etc) because in the first case there is `p`

mlmoravek:
Ideas are transferred to #906

PR #377, opened 2022-06-27
fix datetimepicker min/max time

joeyberkovitz:
Datetimepicker min/max restrictions didn't work properly with the time selector:

- The hour/minute restrictions applied for all dates, where it would only make sense to restrict the hours/minutes if the date portion of the local datetime matches the min/max date
- The meridiem selector was never restricted, so I adjusted it to disable AM/PM if all hours would be disabled
- The docs-next package threw an error about referencing `_inputItems.scss`, so I corrected the case there. Similarly, the existing docs markdown files were named `inputItems.md`, but the auto-generated files were `inputitems.md`, so I deleted the old ones
- As I'm on the latest Node LTS (16), it auto updated the package lock to V2, although I can revert that change if there's some reason to stick with a now outdated version of Node
- Since I wasn't sure of a good way to test that the time selector works properly, I added a min/max date selector to the docs-next package, although I didn't make the same change on the Vue2 docs


joeyberkovitz:
@jtommy - any chance you were able to review this?

jtommy:
It looks ok but please revert changes on node version because at the moment it runs with node 12/14.

About docs-next, are you able to see the new example?

joeyberkovitz:
reverted the package changes to node 12 and tested there
I was able to see docs-next during dev, although the command `build:docs-next:watch:oruga` fails because I don't have `concurrently`, so I had to run `npx vue-docgen --watch` and `npx vitepress-dev` in separate terminals to make it work

after downgrading to Node 12, Vitepress fails, so presumably it doesn't support Node 12, maybe a good reason to upgrade to the current Node LTS :)

Issue #386, opened 2022-07-21
Is There a Proper Way to Use in SFC or by Manually Imports?


YDX-2147483647:
### Description

[In the doc](https://oruga.io/documentation/#setup), Oruga is installed (universally) by `app.use(some_plugin)` even for individual components.

AFAIK, there’s no `app` in [SFC](https://vuejs.org/api/sfc-spec.html)… Is their a proper way to manually import?

```vue
<script>
  import { OButton } from '@oruga-ui/oruga-next'
  // (not real)

  export default {
    components: {
      OButton
    }
  }
</script>

<template>
    <o-button>Click Me</o-button>
</template>
```

### Why Oruga need this feature

- Already implemented!

  (in a trickier way)

  ```js
  import { OButton } from '@oruga-ui/oruga-next/src/components/button'
  ```

- Easier to migrate from other libraries that support this feature. (Eg. [Element Plus](https://element-plus.org/en-US/guide/quickstart.html#manually-import), [Naive UI](https://www.naiveui.com/en-US/os-theme/docs/usage-sfc))

- (For me :-) Oruga offers unique components (eg. `Inputitems` with autocomplete in Vue 3). For some reason, I use Oruga inside a component rather than through out the whole app. Thanks.

> Why does `from '@oruga-ui/oruga-next'` not work?
>
> In [`src/components/button/index.ts`](https://github.com/oruga-ui/oruga/blob/769c98d7b2a60d102409d089576c9a688cfb5382/packages/oruga-next/src/components/button/index.ts), both `default` (the plugin) and `OButton` (the component) are exported.
>
> ```js
> export default {…} as Plugin
> 
> export {
> Button as OButton
> }
> ```
>
> However in [`src/components/index.ts`](https://github.com/oruga-ui/oruga/blob/769c98d7b2a60d102409d089576c9a688cfb5382/packages/oruga-next/src/components/index.ts), only `default` is imported.
>
> ```js
> import Button from './button'
> ```

### Possible Solutions

- Do nothing — just add `import { OButton } from '@oruga-ui/oruga-next/src/components/button'` into the doc.

- Add another `index.ts`, `export { OButton } from './components/button'`.

  > Therefore we can import several components in one line.
  >
  > ```js
  > import { OInputitems, OField, OSwitch } from '@oruga-ui/oruga-next/components'
  > ```
  >
  > Current tricky version:
  >
  > ```js
[...]

PR #397, opened 2022-08-07
feat(oruga, oruga-next): Export individual components

ishitatsuyuki:
It allows components to be registered locally instead of globally (see [1]).
Some IDE can automatically insert these imports which greatly reduces the hassle
of using tree shaking.

It doesn't handle CSS imports yet, but the old implementation did not either.
A promising solution is to use a bundler plugin, which is adopted by Element
Plus [2].

I'm keeping the old, plugin based API for backward compatibility. Eventually
we can remove it in favor of direct imports.

[1]: https://vuejs.org/guide/components/registration.html#local-registration
[2]: https://github.com/element-plus/unplugin-element-plus

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,60 +70,69 @@ Latest ✔ | Latest ✔ | 10+ ✔ | Latest ✔ | 6.1+ ✔ | IE 11  ✔ |
 
 #### Vue 2
 
-Install Oruga
+1. Install Oruga.
 
 ```bash
 npm install @oruga-ui/oruga
 ```
 
-Then import the full bundle
+2. Import the components:
 
-```js
-import Vue from 'vue';
-import Oruga from '@oruga-ui/oruga';
-import '@oruga-ui/oruga/dist/oruga.css';
+ - To get started quickly, use `Oruga` to register all components:
 
-Vue.use(Oruga);
-```
-or individual components (tree shaking)
+    ```js
+    import Vue from 'vue';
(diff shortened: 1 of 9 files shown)


YDX-2147483647:
Another benefit comparing to suggestions in #386: **Type declarations**.

`import { … } from '@oruga-ui/oruga-next/dist/esm/…` fires ts(2307). (However it just works.)

```
Cannot find module '@oruga-ui/oruga-next/dist/esm/button' or its corresponding type declarations. ts(2307)
```

<details>
    <summary>To reproduce</summary>
    <ol>
        <li>Clone the fork and checkout to the branch.</li>
        <li> <code>npm run build:lib</code> in <code>oruga/packages/oruga-next</code>.</li>
        <li>Then make a new TypeScript Vue project with <code>npm init vue@latest</code> (vue 3.2.37 in fact).</li>
        <li>Finally, <a href='https://docs.npmjs.com/cli/v8/commands/npm-install#description:~:text=a%20folder%20containing%20a%20program%20described%20by%20a%20package.json%20file'><code>npm install</code> locally</a> (no error) or from npmjs.com (ts 2307).</li>
    </ol>
</details>

jtommy:
@ishitatsuyuki thanks

@YDX-2147483647 thanks for the point

ishitatsuyuki:
btw, you might want to do a release since the README was updated in this PR

PR #400, opened 2022-08-10
Handle non existing window.matchMedia

tcitworld:
## Proposed Changes
Handle `window.matchMedia` being undefined properly. This can happen when it's hard to mock environment (jest/jsdom) like done in d0262fd675f51d5262737ec1ec3fc0f481db776b.

This should be harmless.


jtommy:
And oruga-next?

tcitworld:
Of course :see_no_evil:

jtommy:
@tcitworld you should remove `?` char because the build doesn't work as you can see

jtommy:
@tcitworld any news?

PR #413, opened 2022-09-17
Fix/imports not used as values

daniil4udo:
## Issue
Attempt to use **oruga-next** with vite@3 failing due to `Uncaught SyntaxError: ambiguous indirect export: App`

## Proposed Changes

- split type and value imports
- add `importsNotUsedAsValues` to the tsconfig.json

### DIFF
--- a/packages/oruga-next/src/components/autocomplete/index.ts
+++ b/packages/oruga-next/src/components/autocomplete/index.ts
@@ -1,4 +1,4 @@
-import { App, Plugin } from 'vue'
+import type { App, Plugin } from 'vue'
 
 import Autocomplete from './Autocomplete.vue'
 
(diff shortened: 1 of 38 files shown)


jtommy:
Why does it throws a similar error?

daniil4udo:
It does? I am using current branch as a temp fix, it runs without any errors now. 

To with error are you referring exactly?

jtommy:
The issue using vite 3. Docs next use vitepress and vite without any issues

daniil4udo:
I assume because I am importing components directly from `src` folder

jtommy:
And why do you need it?

daniil4udo:
In order to use `unplugin-auto-import` that's the only way - to import components directly from **src** folder.

Although I don't get your concern, I am not trying to change a logic or API, this PR simply an improvement, and good practice.

jtommy:
@daniil4udo you are right, thanks!

Issue #417, opened 2022-09-22
Datepicker - month picker differences in prop types


beljand:
### Overview of the problem

**Oruga** version: [0.5.6]
**Vuejs** version: [2.7.8]
**OS/Browser**: MacOS / Chrome

### Description

`selectabledates` in https://github.com/oruga-ui/oruga/blob/develop/packages/oruga/src/components/datepicker/Datepicker.vue#L378 must be and `Array`, but when using a Month picker, the same prop can be also a `Function` - https://github.com/oruga-ui/oruga/blob/develop/packages/oruga/src/components/datepicker/DatepickerMonth.vue#L61.

Vue throws a warning `Invalid prop: type check failed for prop "selectableDates". Expected Array, got Function`

### Steps to reproduce

- use `ODatepicker`
- add `type="month"`
- provide a function for `selectableDates`

### Expected behavior

- prop `selectableDates` in `Datepicker.vue` should accept both `Array | Function`

### Actual behavior

- prop `selectableDates` in `Datepicker.vue` accepts only `Array` while `DatepickerMonth` accepts `Array | Function`

jtommy:
You are right, can you make a PR?

beljand:
> You are right, can you make a PR?

Will do

PR #418, opened 2022-09-26
Fix/datepicker props type

beljand:
Fixes https://github.com/oruga-ui/oruga/issues/417

## Proposed Changes

- `unselectableDates` and `selectableDates` should also accept `Function` so it's compatible with `ODatepickerMonth`


jtommy:
And oruga-next?

jtommy:
@beljand  And oruga-next?

beljand:
> @beljand And oruga-next?

Sorry, forgot about `next`...

PR #312, opened 2022-03-09
feat: vue 3 testing

astagi:
## Proposed Changes

- Cypress component testing for Oruga Next

PR #420, opened 2022-09-28
fix: make controls respond to the `invalid` event

### LINKED ISSUE #407: FormElementMixin should listen for the `invalid` event
### Overview of the problem

**Oruga** version: 0.5.5
**VueJS** version: 3.2.33 (probably affects 2.x as well)
**OS/Browser**: (all)

### Description

`FormElementMixin` checks for HTML5 constraint validation errors on blur and updates the parent field's message if a constraint fails. This works just fine if the constraint failure was caused by user input. However, this doesn't account for cases where the browser itself triggers constraint validation, such as submitting a form.

It seems like the ideal solution would be to listen for `invalid` events and call `setValidity` in the event handler rather than in `checkHtml5Validity`.

### Steps to reproduce

1. Add a required `OInput` to a form:
  ```vue
  <form>
    <OField label="Name">
      <OInput required />
    </OField>
    <OButton>Submit</OButton>
  </form>
  ```
2. Submit the form without ever focusing the name field.

### Expected behavior

Constraint validation should prevent the form from being submitted, and the field should display an error message matching the failed validation constraint (e.g. "This field is required"), just like it would if you simply focused and blurred the element.

### Actual behavior

The browser prevents form submission, and it focuses the empty field and displays its own constraint validation pop-up, but the `OField` doesn't display a message until the field loses focus.

blm768:
Rather than emitting `invalid` events on blur, we should be listening for them.

Fixes #407

## Proposed Changes

- `FormElementMixin.checkHtml5Validity` checks the `validity` property rather than calling `checkValidity`, so it doesn't emit `invalid` events on blur.
- Controls that use `FormElementMixin` respond to `invalid` events, which the browser fires when their containing form is submitted.

## Open Questions

- [ ] Should we suppress the browser's native HTML5 validation pop-ups?
  When the form is submitted, most browsers will focus the first invalid field and display a pop-up with a validation message. This is standard behavior for plain HTML inputs, but for Oruga controls inside a field, the pop-up is redundant because we provide our own user-visible error message. Should we suppress the browser's native pop-up by default?

### DIFF
--- a/packages/docs-next/components/Autocomplete.md
+++ b/packages/docs-next/components/Autocomplete.md
@@ -73,6 +73,7 @@ title: Autocomplete
 | blur              |                                                                              |
 | focus             |                                                                              |
 | update:modelValue |                                                                              |
+| invalid           |                                                                              |
 | select            | **selected** `Object` - selected option<br/>**event** `Event` - native event |
 | infinite-scroll   |                                                                              |
 | typing            |                                                                              |
--- a/packages/docs-next/components/Datepicker.md
+++ b/packages/docs-next/components/Datepicker.md
@@ -91,6 +91,7 @@ title: Datepicker
 | range-end         |            |
 | blur              |            |
 | focus             |            |
+| invalid           |            |
 | update:modelValue |            |
 | change-month      |            |
 | change-year       |            |
--- a/packages/docs-next/components/Datetimepicker.md
+++ b/packages/docs-next/components/Datetimepicker.md
@@ -68,6 +68,7 @@ title: Datetimepicker
 | change-year       |            |
 | blur              |            |
 | focus             |            |
+| invalid           |            |
 | update:modelValue |            |
 
 ## Slots
--- a/packages/docs-next/components/Input.md
+++ b/packages/docs-next/components/Input.md
@@ -56,6 +56,7 @@ title: Input
 | ----------------- | ---------- | ----------- |
 | blur              |            |
 | focus             |            |
+| invalid           |            |
 | update:modelValue |            |
 | icon-click        |            |
 | icon-right-click  |            |
--- a/packages/docs-next/components/Inputitems.md
+++ b/packages/docs-next/components/Inputitems.md
@@ -71,6 +71,7 @@ title: Inputitems
 | icon-right-click  |            |
 | blur              |            |
 | focus             |            |
+| invalid           |            |
 | update:modelValue |            |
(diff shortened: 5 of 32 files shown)


jtommy:
> Should we suppress the browser's native HTML5 validation pop-ups?
> When the form is submitted, most browsers will focus the first invalid field and display a pop-up with a validation message. This is standard behavior for plain HTML inputs, but for Oruga controls inside a field, the pop-up is redundant because we provide our own user-visible error message. Should we suppress the browser's native pop-up by default

Yes we can avoid it

jtommy:
@blm768 the PR looks ok but probably it's better release it in the next breaking version (0.6.0), what do you think?

blm768:
Waiting until 0.6.0 should be OK; I've got a workaround that I can use for this in the meantime. That'd also give us more time to evaluate suppressing the browser's native pop-up messages. We'll probably only want to do that for controls with a parent field, and it might be worthwhile to make it configurable, although the existing `useHtml5Validation` flag might be good enough.

astagi:
@blm768 in the meantime check some github comments about linting :) is it ready for review? I can mark it as breaking and remind to merge in the next version as @jtommy suggested! Let me know!

astagi:
@blm768 the work in [this](https://github.com/oruga-ui/oruga/pull/312) branch is going to finish, we're ready to review all the breaking changes :)

About your PR, lgtm and I also agree with you that Oruga should not come with extra dependencies (polyfills included). I don't know which is the best option, seems ok for me to use `scrollIntoViewIfNeeded` only on browsers that support it and just focus the input on others. The third option (handler function defined by users) looks also good, do you mean to make users able to define their own handler function alongside a standard behaviour (option 1 & 2)?

@jtommy any thoughts?

blm768:
> The third option (handler function defined by users) looks also good, do you mean to make users able to define their own handler function alongside a standard behaviour (option 1 & 2)?

Yep; that's what I had in mind. We'd probably use the standard behavior if the user doesn't provide a handler function and skip our built-in behavior entirely if the user does provide one.

jtommy:
> > The third option (handler function defined by users) looks also good, do you mean to make users able to define their own handler function alongside a standard behaviour (option 1 & 2)?
> 
> Yep; that's what I had in mind. We'd probably use the standard behavior if the user doesn't provide a handler function and skip our built-in behavior entirely if the user does provide one.

I agree with you!

PR #423, opened 2022-10-09
Fix modal calling onClose twice

### LINKED ISSUE #333: Modal onClose method called twice
### Overview of the problem

```
Vue: 3.2.25
@oruga-ui/oruga-next: 0.5.4
@oruga-ui/theme-bulma: 0.2.3
```

### Description

The Modal onClose functions seems to call twice. This behaviour makes it difficult to handle modal callbacks to the originating component.

### Steps to reproduce

[CodeSandbox](https://codesandbox.io/s/oruga-vue3-modal-test-0hcqxy?file=/src/App.vue)

### Expected behavior

Emitting a single 'close' event should only call the `onClose` method a single time.

### Actual behavior

The `onClose` method is called twice, one of which is always with an undefined result in the callback.

travatomic:
Fixes #333

In `oruga-next`, the `close` event was being sent twice. Both `this.$emit('close')` and `this.onClose.apply(null, arguments)` send the `close` event. We can delete either of those statements to get a single `close` event to fire. I chose to delete `this.$emit('close')` because it wasn't sending arguments.

This double `close` event behavior does not seem to happen in `oruga`, so I did not make any changes there.

### DIFF
--- a/packages/oruga-next/src/components/modal/Modal.vue
+++ b/packages/oruga-next/src/components/modal/Modal.vue
@@ -301,7 +301,6 @@ export default defineComponent({
             if (this.destroyOnHide) {
                 this.destroyed = true
             }
-            this.$emit('close')
             this.$emit('update:active', false)
             this.onClose.apply(null, arguments)
 


jtommy:
The oruga and oruga-next code is the same.
Your PR doesn't look correct because the modal won't emit the close event.

travatomic:
@jtommy This appears to be a difference between vue2 and vue3. In vue3, if there is a function prop called `onSomething`, and we execute `this.onSomething.apply`, it will emit the `something` event. That's why we are seeing the `close` event is getting emitted twice in oruga-next, but only once in oruga.

I have made an example to illustrate this in vue3 outside of the  context of the oruga project:
https://codesandbox.io/s/nervous-rgb-zs2dke?file=/src/components/HelloWorld.vue
HelloWorld executes both `this.$emit("foo")` and `this.onFoo.apply(null, arguments)`
App.vue subscribes via `@foo` and shows an alert. If you click the "hello world" button you will see the alert fires twice, signifying that both `this.$emit("foo")` and `this.onFoo.apply(null, arguments)`

So, if we want the event to only fire once, we need to delete either `$emit('close')` or the `onClose.apply` in oruga-next.

jtommy:
Really?? And it's an official Vue 3 behavior?

travatomic:
I'm not sure if that was the Vue team's intention, I'm just reporting what I'm seeing as the cause for #333 😄 . I'm pretty new to Vue, so no doubt you are much more familiar with it. Happy to make changes if you see a better solution.

BTW I was able to reproduce #333 in the oruga-next docs application in the develop branch. I added a `@close=close` to the first modal [here](https://github.com/oruga-ui/oruga/blob/b6870bb8da0ffc56f9544bc460a69a2d5b8a4e2c/packages/docs-next/.vitepress/theme/examples/modal/base.vue#L20) and a close() function that simply executes console.log().

Additionally there are multiple ways to open modals and get close events, so it's possible I've overlooked something, but this solution did not seem to break any of the examples in the docs application, or the `@close=` callback.

travatomic:
@jtommy I'm not sure if that was the Vue team's intention, I'm just reporting what I'm seeing as the cause for #333 😄 . I'm pretty new to Vue, so no doubt you are much more familiar with it. Happy to make changes if you see a better solution.

BTW I was able to reproduce #333 in the oruga-next docs application in the develop branch. I added a `@close="close"` to the first modal [here](https://github.com/oruga-ui/oruga/blob/b6870bb8da0ffc56f9544bc460a69a2d5b8a4e2c/packages/docs-next/.vitepress/theme/examples/modal/base.vue#L20) and a close() function that simply executes console.log().

Additionally there are multiple ways to open modals and get close events, so it's possible I've overlooked something, but this solution did not seem to break any of the examples in the docs application, or the `@close=` callback.

travatomic:
@jtommy What do you want to do here? It seems like either `this.$emit('close')` or `this.onClose.apply(null, arguments)` needs to be removed in oruga-next if we want this callback to fire only once. Unless you see something I don't. 😄 

Happy to make any changes, or set up the docs sites to demonstrate the issue if that would help. (IIMO, I don't think the close callback is important enough to be a permanent addition to the docs site though).

jtommy:
@travatomic probably in vue 3 event are passed as props with on[EventName]

PR #425, opened 2022-10-09
Fix Steps docs

travatomic:
When I created #422, I did not realize I was editing a generated file (`packages/docs/components/Steps.md`). This PR adds that code to the source file (`packages/oruga/src/components/steps/examples/Steps.md`).

As far as I can tell, I updated the `oruga-next` docs correctly in #422. But can you confirm that for me?


jtommy:
No problem, thanks

PR #433, opened 2022-10-25
Fixes message slot in OField component when horizontal

### LINKED ISSUE #427: OField message slot not rendered when horizontal
### Overview of the problem

**Oruga** version: [0.5.6]
**Vuejs** version: [2.7.10]
**OS/Browser**:

### Description

When using 
``<o-field horizontal>
``
`         <template #message>Message goes here</template>`
`
  </o-field>`

The #message slot is not rendered.

### Steps to reproduce

1. Create an o-field with `horizontal` set to `true`
2. add child template for slot #message

### Expected behavior

Message slot should be rendered

### Actual behavior

Message slot is not rendered

yoji-kojio:
Fixes #427 

## Proposed Changes

- Keep `message` slot when component has `horizontal` props.
- Refactor `Field` component to avoid code duplication.
- Fix input spacing on mobile.

Before:
image
image

After:
image
image

### DIFF
--- a/packages/docs-next/.vitepress/theme/examples/field/inspector.vue
+++ b/packages/docs-next/.vitepress/theme/examples/field/inspector.vue
@@ -74,6 +74,10 @@ export default {
           }
         },
         {
+          class: "fieldInputWrapperClass",
+          description: "Class for wrapper that contains input and message",
+        },
+        {
           class: "bodyClass",
           description: "Class for field body"
         },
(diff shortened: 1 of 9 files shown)


yoji-kojio:
@jtommy this PR needs you to approve running workflows :smiley_cat:

jtommy:
It looks ok but probably it's a breaking change. I need to try to understand if we can publish now or in the next 0.6.0 release

yoji-kojio:
> It looks ok but probably it's a breaking change. I need to try to understand if we can publish now or in the next 0.6.0 release

ok! no problem! 😄
I changed a little bit of the component structure, but for the user the only difference will be the message that will be showed correctly on horizontal OField!

PR #450, opened 2022-11-23
Update Pagination.vue

mmorainville:
## Proposed Changes

- Allow `order` property to be changed using global config properties, to be able to change it in the table component, and also because it can be useful.


jtommy:
@mmorainville and oruga-next?

astagi:
@mmorainville thanks for adding the default! @jtommy wdyt? 

@mmorainville Could you make it available for oruga-next as well?

jtommy:
For me it's ok, waiting for oruga-next part

PR #470, opened 2022-12-29
fix: changed background for background-color on .o-chk_check

area73:
Fixes #

## Proposed Changes

- change CSS rule on `o-chk__check`, from:

 `@include avariable('background', 'checkbox-background-color', $checkbox-background-color);` 
to 

`@include avariable('background-color', 'checkbox-background-color', $checkbox-background-color);`

## Context

Right now we are using the `background` shorthand  to set a background color on ourga checkbox:

image

This will override any other settings made in background rule (like background-position, backgrouns-size, etc) and previously set on the same class

This property should be `background-color` instead:

image

### DIFF
--- a/packages/oruga/src/scss/components/_checkbox.scss
+++ b/packages/oruga/src/scss/components/_checkbox.scss
@@ -1,4 +1,3 @@
-
 /* @docs */
 $checkbox-active-background-color: $primary !default;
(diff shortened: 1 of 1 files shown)


jtommy:
@area73 thanks 👍

PR #476, opened 2023-01-21
fix: tabbed child disappearing issue

astagi:


jtommy:
@urkle why not in the theme?

jtommy:
@astagi why not in the theme?

jtommy (commented):
@astagi why not in the theme?

astagi:
@jtommy don't remember what I did, will check it out tomorrow :) moving fix on tue theme may have sense

PR #485, opened 2023-02-09
Update Table.vue to fix footer rendering

luistrcastro:
## Issue

- hasCustomFooterSlot as a function may be called when $slots.footer is not yet available and returns false. 

## Proposed Changes

- hasCustomFooterSlot should be a computed property, because it needs to recompute if $slots.footer changes

ps: actually, I don't get why Oruga tries to asses if it's a custom footer since there's no default footer. Just have the slot there and let users do whatever they want. Maybe bind columnCount to the slot and call it a day.

### DIFF
--- a/packages/oruga-next/src/components/table/Table.vue
+++ b/packages/oruga-next/src/components/table/Table.vue
@@ -266,7 +266,7 @@
 
                 <tfoot v-if="$slots.footer">
                     <tr :class="footerClasses">
-                        <slot name="footer" v-if="hasCustomFooterSlot()"/>
+                        <slot name="footer" v-if="hasCustomFooterSlot"/>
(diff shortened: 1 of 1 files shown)


jtommy:
Are you sure that computed listens $slots changes?

mlmoravek:
@luistrcastro I will add column-count and row-count as props to the footer slot. 

I agree with @jtommy. I don't think a calculated property can watch if anything changes inside a slot. However, the `v-if="$slots.footer"` which is already in the code ensures that the `hasCustomFooterSlot` function is called first after a slot exists.

The `hasCustomFooterSlot` function not only checks if the footer slot is present, it also checks the html tag type of the first element of the slot, so that the slot can be wrapped by a `th` or not if the slot already provides the `th`.

PR #523, opened 2023-05-05
feat: add the clearIcon prop to input and autocomplete

fabiiomiiguel:
Fixes #

Not possible to change the default icon of a clearable input/autocomplete.

## Proposed Changes

- Add a prop to change the default clear icon


packages/oruga-next/src/components/autocomplete/Autocomplete.vue
```diff
@@ -156,6 +156,13 @@ export default defineComponent({
         keepOpen: Boolean,
         /** Add a button/icon to clear the inputed text */
         clearable: Boolean,
+        /**
+         * Icon name to be added on the clear button
+         */
+        clearIcon: {
+            type: String,
+            default: 'close-circle'
```
jtommy, line 164: Can you use the syntax that read it from global config in order to change it in the initial config?

jtommy (changes requested):

jtommy (approved):

PR #528, opened 2023-05-12
better support o-input with input type=number

### LINKED ISSUE #526: o-input with v-model.number handles backspace oddly on 0,1
### Overview of the problem

**Oruga** version: 0.6.0
**Vuejs** version: 3.3.1
**OS/Browser**: Chrome 112

### Description

`o-input` with `v-model.number` doesn't allow to edit numbers freely.

If the input value is `0,1|` (`|` denotes the cursor position) and the user hits backspace, the value is replaced with `|0`.

Native input (which `o-input` is supposed to proxy) in the same situation changes to `0,|`.

This is very annoying when editing numbers quickly.

### Steps to reproduce

```vue
<script setup>
const amount = ref(0.1)
</script>

<template>
  <o-input type="number" step="0.001" v-model.number="amount" />
</template>
```

https://stackblitz.com/edit/vue-aawbbj?file=src%2FApp.vue

1. Focus oruga input, click backspace.
2. Compare to how more naturally native input behaves below.

### Expected behavior

`o-input` and `input` handle editing keystrokes identically.

### Actual behavior

`o-input` behaves unnaturally and doesn't allow to edit text freely. `0,1` replaces with `0` on backspace, and cursor position jumps.

### LINKED ISSUE #527: o-input with type=number should set v-model.number automatically
### Description

Same as native Vue inputs, `o-input` should automatically apply the `number` modifier if the input has `type="number"`.

Currently, `<o-input type="number" v-model="value">` emits value as a string.

[Vue documentation](https://vuejs.org/guide/essentials/forms.html#number) says:

> The `number` modifier is applied automatically if the input has `type="number"`.

It should be straightforward to replace `<input>` with `<o-input>` and non-applying `number` modifier is a gotcha. Together with #526 editing numbers is annoying both for the developer and the user.

IlyaSemenov:
Fixes #526
Fixes #527

## Proposed Changes

This fixes `o-input` with `type=number` (see the linked issues).

Vue doesn't support `modelModifiers` fallthrough to native inputs (see https://github.com/vuejs/core/issues/1825) but simply converting the code to use `v-model` in `o-input` achieves the desired behaviour **and** makes it more consistent with native input.

Besides, the code becomes cleaner. I mean, why introducing a settable computable and not use?

### DIFF
--- a/packages/oruga-next/src/components/input/Input.vue
+++ b/packages/oruga-next/src/components/input/Input.vue
@@ -8,8 +8,7 @@
             :type="newType"
             :autocomplete="newAutocomplete"
             :maxlength="maxlength"
-            :value="computedValue"
-            @input="onInput"
+            v-model="computedValue"
             @blur="onBlur"
             @focus="onFocus"
             @invalid="onInvalid" />
@@ -20,8 +19,7 @@
             ref="textarea"
             :class="inputClasses"
             :maxlength="maxlength"
-            :value="computedValue"
-            @input="onInput"
+            v-model="computedValue"
             @blur="onBlur"
             @focus="onFocus"
             @invalid="onInvalid"
@@ -303,10 +301,6 @@ export default defineComponent({
             this.$nextTick(() => {
                 this.focus()
             })
-        },
-
-        onInput(event) {
-            this.computedValue = event.target.value
         },
 
         iconClick(emit, event) {


jtommy:
> Besides, the code becomes cleaner. I mean, why introducing a settable computable and not use?

Maybe due to integration with other libs like imask or similar

IlyaSemenov:
Note that this change will somehow break mobile native datepicker (`<o-input type="date" @update:modelValue>` will be called with old values somehow).

PR #543, opened 2023-06-13
[Autocomplete] Fix issue related to select-header and select-footer events

andrzejkupczyk:
I have made some changes that address the following issues:

1) Missing events in the `emits` array
2) Not triggering `select-footer` event

Please note that I have only tested these changes manually. If you would like me to incorporate any modifications into the automatic tests, please let me know.

Additionally, this is only applicable to the `oruga-next` package. The `oruga` package looks fine to me: https://github.com/oruga-ui/oruga/blob/develop/packages/oruga/src/components/autocomplete/Autocomplete.vue#L568


jtommy:
@andrzejkupczyk thanks!!

Issue #502, opened 2023-03-28
DropdownItem tag="nuxt-link" does not work


thedomeffm:
### Overview of the problem

**Oruga** version: [0.5.10]
**Vuejs** version: [3]
**OS/Browser**: Chrome

### Description

The `tag="router-link"` tag does work, but `tag="nuxt-link"` does not.

```vue
  <o-dropdown v-if="row.isPaid"
              position="bottom-left"
              aria-role="list">
      <template #trigger="{ active }">
          <button class="button is-ghost has-text-black">
              <o-icon icon="ellipsis-vertical" />
          </button>
      </template>

      <o-dropdown-item tag="router-link"
                       :to="localePath({ name: 'e-id', params: { id: row.code } })"
                       aria-role="listitem">
          <o-icon icon="up-right-from-square" />
          <span>{{ t('table.action.preview') }}</span>
      </o-dropdown-item>

      <o-dropdown-item tag="nuxt-link"
                       :to="localePath({ name: 'e-id', params: { id: row.code } })"
                       aria-role="listitem">
          <o-icon icon="up-right-from-square" />
          <span>{{ t('table.action.preview') }}</span>
      </o-dropdown-item>
  </o-dropdown>
```

Will result in the following **rendered** HTML:

![image](https://user-images.githubusercontent.com/29406401/228347175-0dd8555e-59f7-4c3f-b386-a0e34b0d64ba.png)

### Expected behavior

The `nuxt-link` as tag should work like the `router-link`. (nuxt does understand my current workaround that I use `router-link`).

### Actual behavior

See screenshot - the first link (with tag="router-link") gets rendered correct.
The second drop-down-item (with tag="nuxt-link") gets **rendered** as.... a `<nuxt-link>` and not a `<a>`.

thedomeffm:
@jtommy we wrote in Discord about the problem already.
You send me the link => https://github.com/nuxt/nuxt/issues/13659

So it seems like it is the "cOrReCt" behavior of nuxt3 :/

fahdaddi:
@jtommy, still having the same issue when using o-button with `tag="nuxt-link"`...

PR #548, opened 2023-06-18
Add components to tag prop types

jonathanjameswatson:
Fixes warnings in #502

## Proposed Changes

- Adds `Component` to the types of all tag props, allowing actual components to be given as tag props, in addition to component names.


jtommy:
@jonathanjameswatson thanks!!

PR #551, opened 2023-06-22
Fix Table Pagination Size

### LINKED ISSUE #550: Table Pagination Size is not working
### Overview of the problem

**Oruga** version: 0.5.10
**Vuejs** version: 3.2.9
**OS/Browser**: Win 10, Browser

### Description

Pagination Size does not work, props are wrong on Table.vue

robertmarney:
Fixes #550

## Proposed Changes

- Fix Table Props


jtommy:
@robertmarney thanks!

PR #567, opened 2023-08-11
Fix memory leak in MatchMediaMixin

### LINKED ISSUE #566: Match media memory leak
### Overview of the problem
MatchMediaMixin has a memory leak 

**Oruga** version: [0.6.0]
**Vuejs** version: [3.3.4]
**Chrome** version 118.0.5941.0

### Description
MatchMediaMixin references undefined method in beforeUnmount hook (for example [line 43](https://github.com/oruga-ui/oruga/blob/develop/packages/oruga-next/src/utils/MatchMediaMixin.ts#L43) in oruga-next)

The scale of the problem depends heavily on the amount of data you have in scope (e.g. setup function). I can produce numbers like that with just a few route changes
image

### Steps to reproduce
Mount and unmount components with OModal/OTable or others using MatchMediaMixin

https://github.com/oruga-ui/oruga/assets/10117250/30ee7724-edc0-488f-9592-8077c76f6428

setmeright:
Fixes #566

Proposed Changes
- Call correct method for clean up
- Replace lifecycle hooks with browser executed ones
- Store MediaQueryList as a non-reactive reference
- Replace addListener as it's [depricated](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "root",
-  "version": "0.5.10",
+  "version": "0.6.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "root",
-      "version": "0.5.10",
+      "version": "0.6.0",
       "license": "MIT",
       "devDependencies": {
         "conventional-changelog": "^3.1.25",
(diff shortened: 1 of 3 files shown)


jtommy:
@setmeright it looks ok!

PR #605, opened 2023-09-29
feat: update programmatic oruga

mlmoravek:
I updated the programmatic interface for oruga by adding type definitions and usability enhancements. 

**Contains Breaking Changes**

## Proposed Changes

- Added programmatic interface type definitions
- Removed wrapping layer for programmatic interface
- Changed `useProgrammatic` to `useOruga`


jtommy:
I agree with you but it's a breaking change, so we can release it only on next 0.x version

mlmoravek:
> I agree with you but it's a breaking change, so we can release it only on next 0.x version

Yeah, no need to merge it right away. We can wait until we have more for a major release update.

packages/oruga-next/src/components/loading/index.ts
```diff
@@ -3,40 +3,47 @@ import { createVNode, render } from "vue";
...
 import {
     registerComponent,
     registerComponentProgrammatic,
-} from "../../utils/plugins";
-import InstanceRegistry from "../../utils/InstanceRegistry";
+} from "@/utils/plugins";
+import InstanceRegistry from "@/utils/InstanceRegistry";
+
+declare module "@/types" {
+    interface OrugaPrgrammatic {
```
messenjer, line 15: There is this typo `OrugaPrgrammatic` in a few places

mlmoravek: @messenjer I appreciate the review. Typo is fixed :)

PR #706, opened 2023-12-28
docs: Update customisation guidelines

### LINKED ISSUE #705: Tree-shaked Config is not working anymore
### Overview of the problem

**Oruga** version: 0.8.2
**Vuejs** version: 3.3.13

### Description

Documentation states :

> If you use individual imports you can customize each compoment using the Config plugin.

And shows the following example :
```
import { OAutocomplete, OSidebar, Config } from '@oruga-ui/oruga-next';
[...]
```

It is working with version `0.7.0`, but fails in `0.8.2`

### Steps to reproduce

1. Use the documentation code with version 0.7.0 and see that it works
2. Upgrade to 0.8.2 and see the compilation (with Webpack if that matters) failing with the error above

### Expected behavior

Either that the code from the documentation work, or that the documentation show code that works.

### Actual behavior

Compilation fails with  `Module '"@oruga-ui/oruga-next"' has no exported member 'Config'`

121593:
v0.8 changes the Config interface

Fixes #705

## Proposed Changes

- Change `Config` references to `ConfigProgrammatic`
- Remove `menu-class` config that disappeared from the `Autocomplete` component

### DIFF
--- a/packages/docs-next/documentation/customisation.md
+++ b/packages/docs-next/documentation/customisation.md
@@ -67,27 +67,7 @@ createApp(...)
     .use(Oruga, {
         autocomplete: {
             rootClass: 'myautocomplete-root',
-            menuClass: 'myautocomplete-menu',
             itemClass: 'myautocomplete-item',
-            ...
-        }
-    });
-```
-
-If you use individual imports you can customize each compoment using the `Config` plugin.
-
-```js
-import { createApp } from 'vue';
-import { OAutocomplete, OSidebar, Config } from '@oruga-ui/oruga-next';
-
-createApp(...)
-    .component(OAutocomplete)
-    .component(OSidebar)
-    .use(Config, {
-        autocomplete: {
-            rootClass: 'autocomplete-root',
-            menuClass: 'autocomplete-menu',
-            itemClass: 'autocomplete-item',
             ...
         }
     });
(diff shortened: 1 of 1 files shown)


121593:
I'm not sure that these changes are appropriate.

If I understood correctly we only have the global config, or the programmatic config. That's why I removed the 'programmatic config' section.

Let me know if I got it wrong.

mlmoravek:
I don't agree with replacing all config with programmatic, and I wouldn't remove the programmatic section either, because I think a default behaviour while including all components is to set the config via the plugin options parameter. 

The programmatic usage, which is explained in a separate section, is just an extended way of configuring the components.

121593:
Make sense. I updated the PR.
I restored most of it as it was, and instead removed the `Config` section.

Let me know if that's better.

mlmoravek:
@121593 I would change the last example to:
```
import { OAutocomplete, ConfigProgrammatic } from '@oruga-ui/oruga-next';

createApp(...)
    .use(OAutocomplete);

const config = ConfigProgrammatic.getOptions();

const myThemeConfig = {
    ...config,
    autocomplete: {
        rootClass: 'autocomplete-root',
        itemClass: 'autocomplete-item',
        ...
    }
}

ConfigProgrammatic.setOptions(myThemeConfig);
```

However, the global programmatic interface entry `useOruga` should also be mentioned here somewhere, because through `useOruga` every programmatic component is accessible here. However, the whole section should be renamed to `Prorgammatic usage` and expanded. 

But I can do the enhancement later.

mlmoravek:
@121593 I copied your changes and enhance the "programmtic usage" section. Do you agree with my PR #716 ?

121593:
@mlmoravek yes that looks better ! Closing this PR in favor of yours

PR #721, opened 2024-01-04
fix(autocomplete) fix empty slot display and formatter prop usage

### LINKED ISSUE #718: Autocomplete prop "formatter" should not use prop "field" when it has its default value "value"
### Overview of the problem

**Oruga** version: [0.8.2]
**Vuejs** version: [3.3.11]
**OS/Browser**: Windows 10 / Chrome

### Description

To be able to use autocomplete `formatter` prop, the `field` prop needs to be null. But a default value has been added to field prop as "value". So to be able to use the formatter prop, user must explicitely set `:field="null"`. This change happened in October I think, and has not been documented in CHANGELOG.txt.

### Expected behavior

Since the change has not been well documented, breaks existing usage and is hard to track, I expect this is not wanted. I would have expect to use the formatter when a field prop has not been defined.

### Actual behavior

It uses the default value "value", which often leads to an exception in the formatter functions that is mostly used with objects.

### LINKED ISSUE #719: Autocomplete empty slot is not displayed properly
### Overview of the problem

**Oruga** version: [0.8.2]
**Vuejs** version: [3.3.11]
**OS/Browser**: Windows 10/Chrome

### Description

The empty slot template of autocomplete is not displayed properly, the hide animation is triggered immediately instead when there are no results

### Steps to reproduce

On https://oruga.io/components/Autocomplete.html#props, first example:

chrome_fXbmBrPbpr

sylvainpolletvillard:
Fixes #718
Fixes #719

## Proposed Changes

- remove autocomplete prop field default value
- do not hide autocomplete dropdown when no data if an empty slot is defined

### DIFF
--- a/packages/oruga-next/src/components/autocomplete/Autocomplete.vue
+++ b/packages/oruga-next/src/components/autocomplete/Autocomplete.vue
@@ -313,6 +313,14 @@ const emits = defineEmits<{
     (e: "scroll-end"): void;
 }>();
 
+const slots = defineSlots<{
+    header(): any;
+    group(props: { group: object; index: number }): any;
+    default(props: { option: object; value: unknown; index: number }): any;
+    empty(): any;
+    footer(): any;
+}>();
+
 const inputRef = ref<InstanceType<typeof OInput>>();
 const dropdownRef = ref<InstanceType<typeof ODropdown>>();
 const footerRef = ref<HTMLElement>();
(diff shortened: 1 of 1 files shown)


jtommy:
@sylvainpolletvillard thanks!

sylvainpolletvillard:
Oh my bad definitely not intended

PR #722, opened 2024-01-04
fix(taginput): watch TagInput v-model prop with deep in order to trigger update:modelValue event

tcitworld:
Unless the `deep` option is being specified, the `update:modelValue` event is not triggered by the `TagInput` component (only the `add`/`remove` events are).

There may be another way of fixing this, I'm not familiar with this `useVModelBinding` pattern. Other components may be affected by similar issues, I didn't test them.


mlmoravek:
@tcitworld the `useVModelBinding` composable is an adaptation of the [`useVModel()`](https://vueuse.org/core/useVModel/) composable from the vueuse library. Also in Vue 3.4 this may be replaced by Vue natives [`defineModel()`](https://vuejs.org/api/sfc-script-setup.html#definemodel) composable.

PR #725, opened 2024-01-06
fix: convert useComputedClass to defineClass

### LINKED ISSUE #709: Got error "Error: useComputedClass must be called within a component setup function."
### Overview of the problem

**Oruga** version: [0.8.1]
**Vuejs** version: [3.4]
**OS/Browser**:
Mac/Chrome

### Description
When using o-field tag, I see an error.
```
useComputedClass.ts:57 Uncaught (in promise) Error: useComputedClass must be called within a component setup function.
    at useComputedClass(useComputedClass.ts:57:15)
    at Field.vue:235:5
```

It worked fine until 0.8.0, but an error occurs with 0.8.1 and 0.8.2.

### Steps to reproduce

Reproduce with the following tag from the official document.
```
  <section>                                                                                                                                                      
    <o-field label="Name">                                                                                                                                       
        <o-input model-value="Kevin Garvey" />                                                                                                                   
    </o-field>                                                                                                                                                   
  </section> 
```

### Expected behavior

No error occurs

### Actual behavior

Got Error

mlmoravek:
Fixes #709 

Currently, each time a computed is triggered, the `useComputedClass` composable is triggered for each component class. This causes the `getCurrentInstance()` method to be called during the `computed` hook.

I created a new composable called `defineClasses` which will replace the computed component classes and return a reactive `ref` value. This way the `getCurrentInstance()` method got only triggered once synchronous during setup. The class is calculated synchronous during setup too and every time the `suffix` is changed by a watcher. Another watcher is used to change the apply state of the class if given. However, this doesn't require the class to be recalculated. 

This also improves the update time of a computed class, because before, whenever any property or ref value was used to compute the class got updated, the computation for each class calculation used in the computed const got triggered. Now, whenever a property changes, only the relevant class calculation is triggered.

For each component, the "Computed Component Classes" section must be refactored to use the new feature. No component logic or template should be touched.

## Proposed Changes

- refactor computed `useComputedClass` to `defineClasses`
- Update docs to use Vue 3.4 and latest vitepress version
- Extend basic button docs example


mlmoravek:
@darioillusorium Don't use Vue 3.4, use Vue 3.3.x. However, you can also use Oruga 0.7, which is based on the Options API and could work with the latest Vue version. As of 0.8, we have refactored the components to the Composition API, which contains this bug with Vue 3.4.
If you have any other problems, please open a separate issue.

jtommy (approved):

l00k:
When next release with this fix?

mlmoravek:
@l00k 
I want to finish my a11y enhancement first (https://github.com/oruga-ui/oruga/pull/723) - still working on it. 
Then I will see which issues I can fix quickly. 
And then I will do the new release 🙂

I don't want to make any date prediction, but I will try to do it this month - be aware that Oruga is a side project and nobody is working full time on it!

l00k:
I just thought fix (as independent thing) could be pushed without waiting for other bugfixes or improves.

My project which uses oruga is also side thing :) so.. yea

PR #761, opened 2024-02-03
fix(autocomplete): close dropdown on enter or tab key press

### LINKED ISSUE #707: Autocomplete dropdown does not close on enter or tab key press
### Overview of the problem

**Oruga** version: 0.8.2
**Vuejs** version: 3.3.10
**OS/Browser**: Linux Firefox 120.0.1, Linux Chromium 120.0.6099.109

### Description

In Oruga 0.8.x, the dropdown component of the autocomplete control does not close when the user presses the Enter or Tab keys. This is reproducible on the autocomplete documentation page:

https://oruga-ui.com/components/Autocomplete.html

In contrast, previous versions did close when the user presses the Enter or Tab keys; like on the old vue2 documentation page:

https://vue2.oruga.io/components/Autocomplete.html

### Steps to reproduce

1. Navigate to https://oruga-ui.com/components/Autocomplete.html
2. Select "Theme -> Oruga Full" from dropdown at top-left of page (if not already selected)
3. Click the "Find a JS framework" autocomplete input to focus it (dropdown of framework options appears):
   oruga-autocomplete-onfocus
4. Press the "down arrow" key (the first option, "Angular", is highlighted):
   oruga-autocomplete-arrow-down
5. Press the "enter" key

### Expected behavior

The autocomplete input's value is changed to "Angular" -- _and the dropdown disappears_:
oruga-autocomplete-old-enter-key

Pressing the tab key moves the focus to other controls on the page (eventually to the "Find a name" autocomplete input):
oruga-autocomplete-old-tab-key

### Actual behavior

The autocomplete input's value is changed to "Angular" -- _but the dropdown remains visible_:
oruga-autocomplete-enter-key

Pressing the tab key moves the focus to successive options in the dropdown:
oruga-autocomplete-tab-key

### Steps to reproduce

1. Navigate to https://oruga-ui.com/components/Autocomplete.html
2. Select "Theme -> Oruga Full" from dropdown at top-left of page (if not already selected)
3. Click the "Find a JS framework" autocomplete input to focus it (dropdown of framework options appears)
4. Press the "down arrow" key (the first option, "Angular", is highlighted)
5. Press the "tab" key

### Expected behavior

The autocomplete input's value is changed to "Angular" -- _and the dropdown disappears, and the focus moves to the next control_. Pressing the tab key more times eventually moves the focus to the "Find a name" autocomplete input.

### Actual behavior

The autocomplete input's value is changed to "Angular" -- _but the dropdown remains visible, and the focus moves to the first option in the dropdown_. Pressing the tab key more times cycles through the options; the focus never moves to the "Find a name" autocomplete input.

justinludwig:
Fixes #707 with the following changes to the Autocomplete component:

1. Close after selection

    When an autocomplete value is selected via the autocomplete's
    `setSelected()` method, the autocomplete's v-model value will usually be
    updated, and the dropdown will usually scheduled to be closed on the
    next tick. However, updating the v-model value triggers the v-model
    watcher, which would then schedule the dropdown to be open again on the
    next tick.

    This change updates the v-model watcher to not do anything when the new
    value is the same as the currently selected value.

2. Close on blur

    If the autocomplete looses focus without a mouse click (eg because the
    user tabbed away from it), the dropdown would remain open.

    This change extends the standard onBlur handler to first close the
    dropdown (and then run the standard onBlur handling).

3. Remove menu from tab sequence

    Pressing the tab key when the autocomplete input was focused moved the
    focus to autocomplete menu.

    This change removes the autocomplete menu and all of its options from
    the tab sequence.

    This change also adds a `menuTabindex` prop to the Dropdown component
    (to enable the dropdown menu to be removed from the tab sequence).

4. Clear itemRefs on re-render

    Using the up & down keys to change the "hovered" state triggers a
    re-render of the autocomplete menu options -- but not of the
    autocomplete component itself. The autocomplete's own `onBeforeUpdate`
    handler is not called when this happens; and so using it to clear the
    `itemRefs` array resulted in the previous set of options not being
    removed.

    The main manifestation of this bug was that it was not possible to use
    the down key to access a selectable-footer option.

    This change clears the `itemRefs` array whenever the `setItemRef()`
    function is called for the first menu option; now the `itemRefs` array
    will be cleared both whenever the full autocomplete component is
    re-rendered, and whenever just the individual options are re-rendered.

5. Select head/foot exclusively of other options

    When the `selectable-header` (or `selectable-footer` prop) is true, it
    was possible to use the up & down keys to turn on the visual "hovered"
    state for both the header (or footer) option and another regular option
    at the same time.

    This change centralizes the logic of setting the `hoveredOption`,
    `headerHovered`, and `footerHovered` refs in the `setHovered()`
    function, ensuring that only one can be set at a time.

6. Apply ARIA combobox pattern

    The ARIA attributes used by the autocomplete menu and options
    represented the menu as if it were a standalone modal dialog.

    The change applies the Combobox Pattern from the ARIA Authoring
    Practices Guide to the autocomplete input, menu, and options -- which
    should allow assistive technologies to recognize the menu as a list of
    options available for the input, and to identify the currently "hovered"
    option.

    See https://www.w3.org/WAI/ARIA/apg/patterns/combobox/ for details.

    This change also adds a `menuId` prop to the Dropdown component
    (to allow the combobox to reference the dropdown menu by id).


mlmoravek:
@justinludwig Well done PR. Thanks a lot for your effort!!!

PR #768, opened 2024-02-09
Fix: #355 Tabs Animation on clicking previous tab is broken

### LINKED ISSUE #355: Tabs Animation on clicking previous tab is broken
### Overview of the problem

**Oruga** 0.5.5
**Vuejs** 3.2.33
**OS/Browser**:  Chrome 100.0.4896.127

### Description

When clicking the previous tab, no animation is shown for sliding back.

### Steps to reproduce

1. Use the Tabs component

### Expected behavior

A nice slide animation to previous tab should occur

### Actual behavior

it went back to previous tab but without the sliding animation.

mk-startialab:
Fixes #355

## Proposed Changes

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

> The find() method of [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) instances returns the first element in the provided array that satisfies the provided testing function.

So, removed index from result of Array.find() in performAction().

### DIFF
--- a/packages/oruga-next/src/components/tabs/Tabs.vue
+++ b/packages/oruga-next/src/components/tabs/Tabs.vue
@@ -246,7 +246,7 @@ function clickFirstViableChild(startingIndex: number, forward: boolean): void {
 function performAction(newId: number | string): void {
     const oldValue = activeId.value;
     const oldTab = isDefined(oldValue)
-        ? items.value.find((item) => item.value === oldValue)[0]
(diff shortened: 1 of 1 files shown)


mlmoravek:
@mk-startialab I think you have an old version of the default banch. The line of code you want to change is no longer in the repo like this :)

PR #777, opened 2024-02-13
Draft : V8 bulma integration

whiplashwebb:
Makes a few changes to dropdown, input and select to allow the bulma theme to control classes correctly without the need for access to computed or data.

Fixes #

## Proposed Changes

-
-
-


package.json
```diff
@@ -15,6 +15,7 @@
     "build:lib:oruga-next": "lerna run build:lib --parallel --scope @oruga-ui/oruga-next",
     "build:lib:watch:oruga-next": "lerna run build:lib:watch --parallel --scope @oruga-ui/oruga-next",
     "publish:lib:oruga-next": "lerna run publish:lib --scope @oruga-ui/oruga-next",
+    "pack:lib:oruga-next": "lerna run pack:lib --scope @oruga-ui/oruga-next",
```
whiplashwebb, line 18: This is to help with my testing. I'll back this out before merge.

packages/oruga-next/package.json
```diff
@@ -50,7 +50,8 @@
     "test:watch": "rm -rf .nyc_output coverage && NODE_ENV=test cypress open --component",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --quiet --ignore-path .gitignore",
     "gen-volar-dts": "node ../../.scripts/gen-component-declaration.mjs --bundle --platform=node",
-    "update": "ncu -u"
+    "update": "ncu -u",
+    "pack:lib": "echo \"$PWD/$(npm pack)\""
```
whiplashwebb, line 54: This is to help with my testing. I'll back this out before merge.

packages/oruga-next/src/components/select/Select.vue
```diff
@@ -183,6 +184,10 @@ const props = defineProps({
         type: [String, Array, Function] as PropType<ComponentClass>,
         default: undefined,
     },
+    variantClassOnRoot: {
```
mlmoravek, line 187: Why not just add another variant class like `rootVariantClass`? This tweak seems unnecessarily complex.

packages/oruga-next/src/components/input/Input.vue
```diff
@@ -396,6 +400,7 @@ const rootClasses = defineClasses(
         null,
         computed(() => props.expanded),
     ],
+    ["hasIconRightClass", "", null, computed(() => hasIconRight.value)],
```
mlmoravek, line 403: I would suggest something more general like `hasIconClass` with a given suffix like `[left]`, `[right]` or `[left,right]` or something like that. The suffix can also be just a string with one word `left`, `right`, `both` if an array is not suitable.

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -223,6 +224,22 @@ const props = defineProps({
         type: [String, Array, Function] as PropType<ComponentClass>,
         default: undefined,
     },
+    useRootPositionClass: {
```
mlmoravek, line 227: Unnessaray, just provide poth classes if defined. They are always an empty string if not defined and override is true (with the theme has).

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -489,6 +506,20 @@ provideDropdown(provideData);
 
 // --- Computed Component Classes ---
 
+const rootPositionComputedClass: ComputedClass = [
+    "rootPositionClass",
+    "is-",
```
mlmoravek, line 511: Thats not a good default class. Compare to other classes.

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -500,6 +531,14 @@ const rootClasses = defineClasses(
         null,
         computed(() => isMobileModal.value && !hoverable.value),
     ],
+    props.useRootPositionClass ? rootPositionComputedClass : undefined,
+    [
+        "rootActiveClass",
+        "",
```
mlmoravek, line 537: Default class is missing

whiplashwebb: fixed

packages/oruga-next/src/components/select/Select.vue
whiplashwebb: fixed

packages/oruga-next/src/components/tooltip/Tooltip.vue
```diff
@@ -302,6 +306,12 @@ const rootClasses = defineClasses(
         null,
         computed(() => !!props.teleport),
     ],
+    [
+        "rootPositionClass",
```
mlmoravek, line 310: I don't understand the need for this class? By setting `teleport` to true, the actual tooltip HTMLElement is no longer below the root element. As a result, the root element doesn't need to have any visual styling.

packages/oruga-next/src/components/sidebar/Sidebar.vue
```diff
@@ -334,15 +334,18 @@ onBeforeUnmount(() => {
...
 });
 
+let removeOutsideListener = null;
+
 /** add outside click event listener */
 function addHandler(): void {
-    if (isClient && !props.overlay)
-        setTimeout(() => document.addEventListener("click", clickedOutside));
+    if (isClient && !props.overlay) {
+        removeOutsideListener = useClickOutside(rootRef, clickedOutside);
```
mlmoravek, line 342: What does this change have to do with the Bulma integration? Did you fix some other bug here? maybe exclude it in a second PR :)

packages/oruga-next/src/components/select/Select.vue
```diff
@@ -300,6 +304,12 @@ const rootClasses = defineClasses(
         null,
         computed(() => props.expanded),
     ],
+    [
+        "rootVariantClass",
+        "o-ctrl-sel--",
+        computed(() => statusVariant.value || props.variant),
```
mlmoravek, line 310: Could you give me a link why the variant is needed at the root element here?

packages/oruga-next/src/components/input/Input.vue
```diff
@@ -490,6 +507,15 @@ const counterClasses = defineClasses(["counterClass", "o-input__counter"]);
             @invalid="onInvalid"
             @input="onInput" />
 
+        <o-icon
```
mlmoravek, line 510: Changing the HTML structure is a breaking change. Styling could be based on a specific HTML structure. Is it necessary to move the icon element here?

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
```diff
@@ -442,7 +442,7 @@ const isEmpty = computed(
 );
 
 watch(isEmpty, (empty) => {
-    if (isFocused) isActive.value = !empty;
+    if (isFocused.value) isActive.value = !empty || !!slots.empty;
```
mlmoravek, line 445: Nice bug you found here :D Could you exclude it in its own PR? it makes the changelog clearer.

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
```diff
@@ -797,6 +801,10 @@ function itemOptionClasses(option): ClassBind[] {
 
     return [...itemClasses.value, ...optionClasses.value];
 }
+
+defineExpose({
+    focus,
```
mlmoravek, line 806: Hmmm... Maybe we should consider exposing the focus function for all HTML input based components to make it more consistent? In its own PR ;)

packages/oruga-next/src/components/input/Input.vue
```diff
@@ -396,6 +408,12 @@ const rootClasses = defineClasses(
         null,
         computed(() => props.expanded),
     ],
+    [
+        "hasIconRightClass",
```
mlmoravek, line 412: I don't think this class is really necessary, you can check with css if the right icon is enabled with something like ` input, textarea { & + .iconRightClass {...}}`.

packages/oruga-next/src/components/input/Input.vue
```diff
@@ -365,6 +370,14 @@ function rightIconClick(event: Event): void {
 const isPasswordVisible = ref(false);
 const inputType = ref(props.type);
 
+const computedInputType = computed(() => {
+    if (props.passwordReveal) {
```
mlmoravek, line 374: Again, this seems to be better placed in it's own PR :)

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -500,6 +512,19 @@ const rootClasses = defineClasses(
...
+        autoPosition,
+        computed(() => !!autoPosition.value),
+    ],
+    [
+        "rootActiveClass",
+        "o-drop--active",
+        null,
+        computed(() => isActive.value || props.inline),
+    ],
+    ["hoverableClass", "", null, computed(() => hoverable.value)],
```
mlmoravek, line 527: `hoverable` is already a computed prop so `["hoverableClass", "", null, hoverable]` might be better

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -500,6 +512,19 @@ const rootClasses = defineClasses(
...
         computed(() => isMobileModal.value && !hoverable.value),
     ],
+    [
+        "rootPositionClass",
+        "o-drop--position-",
+        autoPosition,
+        computed(() => !!autoPosition.value),
+    ],
+    [
+        "rootActiveClass",
```
mlmoravek, line 522: I think it's fine that the class is just called `activeClass' and `positionClass' without the `root` like the other.

packages/oruga-next/src/components/sidebar/Sidebar.vue
whiplashwebb: I mentioned this in the PR description. There was a unrelated bug where the sidebar closed itself whenever you clicked on just about anything, including the contents of the sidebar. I should be able to break this out.

packages/oruga-next/src/components/tooltip/Tooltip.vue
whiplashwebb: I added this class to replicate how the classes worked in the last version. The [current styles](https://github.com/whiplashwebb/theme-bulma/blob/181835d3d6d55ec7fc95fbaabfa8a6125090796a/src/assets/scss/components/_tooltip.scss#L108) assume the position class will be on the root. It looks like I could update these styles pretty easily to change that assumption, but that seems more risky than replicating the old behaviour. Other devs may have built on top of this assumption and I don't want to disrupt that unless there's a solid reason. 

I didn't consider the impact of the teleport prop. That one is new to me. Wouldn't teleporting the tooltip contents break tooltip positioning? It needs the position:relative on the root to absolutely position the contents. Is there an example of this usage somewhere I can test against?

packages/oruga-next/src/components/select/Select.vue
whiplashwebb: Sure, this is just how Bulma works. See https://github.com/jgthms/bulma/blob/master/sass/form/select.sass#L44

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
whiplashwebb: Yes, this can be broken out cleanly.

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
whiplashwebb: This is something that worked prior to 0.8.0, I found the issue because it broke some customer autofocusing logic we have on our end. I think expanding this to cover all focusable elements would be a good idea, custom focus logic is a common use-case.

packages/oruga-next/src/components/select/Select.vue
mlmoravek: Hmm ok its fine but when I look at all the classes of the select component I think I should refactor them and move some up to the root container element....

packages/oruga-next/src/composables/defineClasses.ts
```diff
@@ -23,7 +23,7 @@ import type {
 } from "@/types";
 
 // named tuple as prop definition
-type ComputedClass = readonly [
+export type ComputedClass = readonly [
```
mlmoravek, line 26: I think the export is no longer needed, isn't it?

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
mlmoravek: Yes, due to the change in component visibility by switching to the composition API.
Please remove it from this PR and make an issue so I can track it more clearly.

packages/oruga-next/src/components/input/Input.vue
whiplashwebb: You're right that it could work like that, but this is another spot where this is just how Bulma works. [See this code](https://github.com/jgthms/bulma/blob/master/sass/form/tools.sass#L171) for details.

whiplashwebb: Actually, now that I look at this again, I need to add a `hasIconLeft` prop as well.

mlmoravek: Maybe then a `hasIconClass` with the suffix like `left`, `right`,  `both` ?

packages/oruga-next/src/components/input/Input.vue
whiplashwebb: Ok I can break it out.

packages/oruga-next/src/components/dropdown/Dropdown.vue
whiplashwebb: Good point, that's redundant. Fixed.

packages/oruga-next/src/components/dropdown/Dropdown.vue
whiplashwebb: Fixed.

packages/oruga-next/src/composables/defineClasses.ts
whiplashwebb: In the end I didn't need it, but I think it's still a good idea to export it so it can be used in other files. That type is used implicitly all over the place and it can be handy for debugging at the very least. At worst it's a few extra bytes in the package which will never make it into anyone's production build.

packages/oruga-next/src/components/input/Input.vue
whiplashwebb: If both are present there needs to be two classes rather than one. The has icon left class is currently handled in the theme because it only needs prop data, while the right class needs access to the computed.

packages/oruga-next/src/components/tooltip/Tooltip.vue
whiplashwebb: I understand what the teleport prop does here, I'm trying to understand the real-world usage of this prop. Is there an example I can look at in the other themes? 

Curiosity about teleport aside, I don't see the harm in adding this class. Your point about having styles on the tooltip content is valid, but we still have the output of `positionClass` in `contentClasses`, so this shouldn't break anything and the class can always be disabled in other themes if it's not desired.

packages/oruga-next/src/components/sidebar/Sidebar.vue
whiplashwebb: Now PR https://github.com/oruga-ui/oruga/pull/786

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
whiplashwebb: Now tracked as https://github.com/oruga-ui/oruga/pull/787

packages/oruga-next/src/components/autocomplete/Autocomplete.vue
whiplashwebb: Now tracked as https://github.com/oruga-ui/oruga/pull/788

packages/oruga-next/src/components/input/Input.vue
mlmoravek: but the theme could handle it with a function based on the given suffix? The function got called every time the suffix change. 

However, I'm also fine with adding two classed one for left and one for right, maybe its more clear.

packages/oruga-next/src/components/input/Input.vue
mlmoravek: Ahh ok, I get it. I changed it because I thought this html structure would be more intuitive. But that's fine with your argument.

packages/oruga-next/src/components/tooltip/Tooltip.vue
mlmoravek: But does it work for people using the Bulma theme and the teleport prop?

For example, if you use teleport, the tooltip will not refer to a container with overflow hidden or something like that, where it would be cut off.

packages/oruga-next/src/composables/defineClasses.ts
mlmoravek: But it will never be needed outside, because we define all defineClasses props directly at the function call. Just a programming paradigm like don't add code that is not used.

packages/oruga-next/src/components/tooltip/Tooltip.vue
whiplashwebb: that's a totally valid point, i really don't know the answer. that's why i'm asking for an example of usage, i don't understand what "working" would even look like with teleport. Bulma does have this feature, neither does Buefy, it's new to Oruga. Is there an example of this in action from the world of bootstrap? i used to do a ton of bootstrap, but it's been years at this point.

whiplashwebb: Within vanilla bulma this just wouldn't work. There's a parent/child relationship intentionally written into the code. We could certainly add something to Oruga/Bulma to support this, but it would be custom code on top of Bulma, and I would argue it should be tackled as a separate feature.

packages/oruga-next/src/components/input/Input.vue
whiplashwebb: Maybe? I can't think of a way to do it with the existing suffix mechanism, but that doesn't mean it's impossible. Either way, I think we should keep it like this for now. It's clear and simple this way.

packages/oruga-next/src/composables/defineClasses.ts
whiplashwebb: fair enough, removed.

packages/oruga-next/src/components/tooltip/Tooltip.vue
mlmoravek: nvm. the root container classes get used to the teleport container too. So no worry. However, i think you can adjust the theme css to use the aktuall normal positionClass prop. I don't see the need for the root one looking at the theme code.

packages/oruga-next/src/components/dropdown/Dropdown.vue
```diff
@@ -223,6 +223,18 @@ const props = defineProps({
         type: [String, Array, Function] as PropType<ComponentClass>,
         default: undefined,
     },
+    positionClass: {
```
mlmoravek, line 226: Please add jsdoc descriptions to all new classes.

whiplashwebb: oops. added.

packages/oruga-next/src/components/tooltip/Tooltip.vue
```diff
@@ -161,6 +161,11 @@ const props = defineProps({
         type: [String, Array, Function] as PropType<ComponentClass>,
         default: undefined,
     },
+    /** Root element class indicated dropdown position */
+    rootPositionClass: {
```
mlmoravek, line 165: I'm not fine with the rootPositionClass on the Tooltip component. I think it is not needed. The position is a feature which is implemented by theme itself and not by the bulma library. So it should be possible to refactor the theme to use the actuall positionClass. The bootstrap theme did the same.

whiplashwebb: fair enough. removed.

mlmoravek (approved):

PR #786, opened 2024-02-20
refactor(sidebar): use useClickOutside() composable instead of custom listener

whiplashwebb:
This fixes a bug found while testing https://github.com/oruga-ui/oruga/pull/777 where the sidebar would close itself, even when the click was coming from its own contents. 

## Proposed Changes

- Refactors sidebar to use the standard `useClickOutside()` composable instead of custom listener code.


packages/oruga-next/src/components/sidebar/Sidebar.vue
```diff
@@ -334,15 +335,18 @@ onBeforeUnmount(() => {
...
 });
 
+let removeOutsideListener = null;
+
 /** add outside click event listener */
 function addHandler(): void {
-    if (isClient && !props.overlay)
-        setTimeout(() => document.addEventListener("click", clickedOutside));
+    if (isClient && !props.overlay) {
+        removeOutsideListener = useClickOutside(rootRef, clickedOutside);
```
mlmoravek, line 343: I think `rootRef` is wrong, you have to use `sidebarContent`.

whiplashwebb: yeah i like that much better. done.

PR #789, opened 2024-02-20
fix(input): update password reveal 

whiplashwebb:
Fixes an issue encountered while testing https://github.com/oruga-ui/oruga/pull/777 where if the password reveal prop was set the password would not be hidden when the component initially loaded. Subsequently clicking the show/hide icon would work correctly.

### DIFF
--- a/packages/oruga-next/src/components/input/Input.vue
+++ b/packages/oruga-next/src/components/input/Input.vue
@@ -365,6 +365,14 @@ function rightIconClick(event: Event): void {
 const isPasswordVisible = ref(false);
 const inputType = ref(props.type);
 
+const computedInputType = computed(() => {
+    if (props.passwordReveal) {
+        return isPasswordVisible.value ? "input" : "password";
+    } else {
+        return inputType.value;
+    }
+});
+
 // update inputType on type prop change
 watch(
     () => props.type,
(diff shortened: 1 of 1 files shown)


packages/oruga-next/src/components/input/Input.vue
```diff
@@ -365,6 +365,14 @@ function rightIconClick(event: Event): void {
 const isPasswordVisible = ref(false);
 const inputType = ref(props.type);
 
+const computedInputType = computed(() => {
+    if (props.passwordReveal) {
+        return isPasswordVisible.value ? "input" : "password";
```
messenjer, line 370: I think it's "text" instead of "input"

whiplashwebb: hahaha i can't believe i did that! good catch! ironically the default value saved me, it's surprising that worked at all. fixed.

packages/oruga-next/src/components/input/Input.vue
```diff
@@ -365,6 +365,14 @@ function rightIconClick(event: Event): void {
 const isPasswordVisible = ref(false);
 const inputType = ref(props.type);
 
+const computedInputType = computed(() => {
```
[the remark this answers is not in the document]
whiplashwebb: that's a really good point. i was even able to work in the watcher. done.

mlmoravek (approved):

PR #812, opened 2024-02-26
fix(select): adjust wrong type for `rootVariantClass` property

whiplashwebb:
## Proposed Changes

- Fixes a mistake I made in the types for select. rootVariantClass should be a ClassDefinition not a boolean. We regret the error.
- Turns out I should have been generating types rather than writing them. This Pr has been a learning experience. I regenerated the types and then patched the output to deal with a possible bug where the types for NotificationNotice were missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,3 @@ packages/docs/.vitepress/.cache/
 packages/docs/.vitepress/.temp/
 packages/docs/.vitepress/dist/
 packages/oruga/README.md
-packages/docs-next/.vitepress/cache/
-packages/docs-next/.vitepress/.cache/
-packages/docs-next/.vitepress/.temp/
-packages/docs-next/.vitepress/dist/
-packages/oruga-next/dist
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -10,3 +10,8 @@ packages/docs/.vitepress/.cache/
 packages/docs/.vitepress/.temp/
 packages/docs/.vitepress/dist/
 packages/oruga/README.md
+packages/docs-next/.vitepress/cache/
```
mlmoravek, line 13: the package name got changed from docs-next to docs, as well as oruga-next to next. So no need for this adding.

packages/oruga/src/components/types.ts
```diff
@@ -865,7 +865,7 @@ but will set body to position fixed, might break some layouts. */
                 /** Class of the select arrow */
                 arrowClass: ClassDefinition;
                 /** Class of the select variant for the root element */
-                rootVariantClass: boolean;
+                rootVariantClass: ClassDefinition;
```
mlmoravek, line 868: did you genereate the file?

.gitignore
whiplashwebb: Clearly I need more coffee. You're right, i just had leftover files in the old directories locally.

packages/oruga/src/components/types.ts
whiplashwebb: oh no... i've been manually editing a generated file... 🤦

PR #832, opened 2024-03-01
fix(table): `stickyHeaderClass` should not display by default

whiplashwebb:
Fixes an issue where the table will display `stickyHeaderClass` when no value is provided for `props.stickyHeader`. 

## Proposed Changes

- Cast `props.stickyHeader` to bool in `tableWrapperClasses`


mlmoravek:
@whiplashwebb Thanks for finding this bug. Maybe better set the prop default to `false`?

whiplashwebb:
@mlmoravek I wasn't sure if the prop was intentionally true/false/undefined so that computed seemed safer, but if you think it's safe i like that much better. fixed.

mlmoravek:
@whiplashwebb please regenerate the docs. The props section should change and the default false should be added.

PR #845, opened 2024-03-07
feat(table): make `bordered`, `striped`, `narrowed` and `hoverable` configurable

whiplashwebb:
Oruga's table is deeply configurable, but with so many options it can be hard for devs to know which combination of options is the standard for a given application. This PR adds config support for a few table props related the visual presentation of the table. This will make it easier to set the visual standard for your app.

## Proposed Changes

- Adds config support for `bordered`, `striped`, `narrowed` and `hoverable` table props, along with related types.

### DIFF
--- a/packages/oruga/src/components/table/Table.vue
+++ b/packages/oruga/src/components/table/Table.vue
@@ -65,7 +65,10 @@ const props = defineProps({
     /** Border to all cells */
     bordered: { type: Boolean, default: false },
     /** Whether table is striped */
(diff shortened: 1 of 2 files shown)


mlmoravek:
@whiplashwebb It's okay to add this. 
I'm totally not a performance expert, but I might want to remember that adding a configurable property adds some computable logic every time the component gets mounted. At some point maybe this can lead to some performance issues having too many config based props.

Issue #596, opened 2023-09-22
Feature: Enhance a11y 


mlmoravek:
### Description
* Add [eslint-plugin-vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility)
* Add ["plugin:vuejs-accessibility/recommended"](https://vue-a11y.github.io/eslint-plugin-vuejs-accessibility/) eslint rule.
* Update all components to fullfill the recommended a11y rules.

### Why Oruga need this feature
Accessibility is a big topic and a component library should implement a11y support by default as good as they can.

mmazanec22:
I'm using Deque's axe for integration and e2e tests on a project that uses oruga (jest-axe for integration, axe via nightwatch for e2e). We recently bumped a few versions from 0.5.10 to 0.8.3. The change created several accessibility problems that jest-axe is catching, including:

- incorrect aria in the steps component
- incorrect aria in the switch component
- nested interactive controls in the switch component

Adding axe to the tests for this project in addition to the recommendations above could help prevent future breaking updates.

mlmoravek:
@mmazanec22 I have no experience with [axe](https://github.com/dequelabs/axe-core). Would you like to make a PR with an example of how to use it? 
However, you are always welcome to add some a11y improvements to the codebase :)

mmazanec22:
@mlmoravek I made an attempt with #846 and fixed an a11y issue in the steps component. If this looks like an ok approach, I'd be happy to continue with the other components that were flagged.

mlmoravek:
@mmazanec22 Since the axe setup is now merged, you are welcome to add more tests and suggest a11y improvements if you can :)

mlmoravek:
@jh0ker You are always welcome to make such a11y suggestions. And I think a new issue for each is better. :)

PR #846, opened 2024-03-07
tests(steps): add axe tests, correct steps aria

mmazanec22:
An example of using [Deque's axe library](https://github.com/dequelabs/axe-core) (via [jest-axe](https://www.npmjs.com/package/jest-axe)) to test accessibility for a component (see #596 ). Also fixes an accessibility issue for the steps component. Modeled after #841, assuming that more tests will be added following this pattern.

## Proposed Changes

- Use jest-axe to test accessibility of the Steps component given different props scenarios
- Remove the [aria-orientation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-orientation) role, as it is not valid on this component
- Add the [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) property


packages/oruga/package.json
```diff
@@ -100,5 +100,8 @@
     "vue-component-meta": "^1.8.27",
     "vue-component-type-helpers": "^1.8.27",
     "vue-tsc": "^1.8.27"
+  },
+  "dependencies": {
+    "jest-axe": "^8.0.0"
```
mlmoravek, line 105: Is it possile to the dependency as dev dependency?

packages/oruga/src/components/steps/tests/StepsExample.vue
```diff
@@ -0,0 +1,127 @@
...
+        default: false,
+    },
+});
+
+const activeStep = ref<number>(1);
+const enableProfileActivateEvent = ref(false);
+
+const onProfileActivate = () => {
+    if (enableProfileActivateEvent.value) {
+        alert("Profile Activated");
```
mlmoravek, line 51: I think using alert here seems wrong. I don't know how it works with a virtual dom?

packages/oruga/src/components/steps/tests/StepsExample.vue
```diff
@@ -0,0 +1,127 @@
...
+};
+</script>
+
+<template>
+    <!-- Taken from the first example in the documentation -->
+    <o-steps
+        v-model="activeStep"
+        :vertical="isVertical"
+        :animated="isAnimated"
+        :rounded="isRounded"
```
mlmoravek, line 62: You only need to specify props for this test component that you will not be using only for the root element. Anything else can be passed to OSteps with `:v-bing=$attrs`.

packages/oruga/src/components/steps/tests/steps.integration.test.ts
```diff
@@ -0,0 +1,59 @@
+import { afterEach, describe, expect, test } from "vitest";
+import { enableAutoUnmount, mount } from "@vue/test-utils";
+import { axe, toHaveNoViolations } from "jest-axe";
+
+import StepsExample from "./StepsExample.vue";
+
+expect.extend(toHaveNoViolations);
```
mlmoravek, line 7: Do we have to add this in every test file now? Or is it a one time think? If so, you can also add a `vitest.setup.ts` file in a `/src/__tests__/` folder and a reference to the setup file in `vite.config.ts` with `test:{ setupFiles: [require.resolve("./src/__tests__/vitest.setup.ts")],}`

packages/oruga/src/components/steps/tests/steps.integration.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        },
+        {
+            title: "axe steps - customNavigation",
+            props: { customNavigation: true },
+        },
+    ];
+
+    test.each(a11yCases)("$title", async ({ props }) => {
+        const wrapper = mount(StepsExample, props);
+        expect(await axe(wrapper.html())).toHaveNoViolations();
```
mlmoravek, line 57: Wow, thats all? nice :D

packages/oruga/src/components/steps/tests/StepsExample.vue
mmazanec22: I copied this from the docs. Happy to remove or replace. I think as long as there is any func here it should satisfy our testing needs

packages/oruga/src/components/steps/tests/StepsExample.vue
mmazanec22: @mlmoravek I tried taking these out but realized: I want to toggle one prop at a time to better isolate a11y problems. In order to do that, I want to set default values in props that stay consistent regardless of what the component's defaults are. What do you think about leaving this as is?

mlmoravek: I think for now its fine :) I can refactor it late on when we will add more steps tests. 
To demonstrate how we can benefit from axe tests its good.

PR #858, opened 2024-03-24
docs: update active text color in theme dropdown

TerrniTLLC:
## Proposed Changes

- Changed the text color of the active element to make it prettier 

Before:
Screenshot_20240324_163624
After: 
Screenshot_20240324_163638


mlmoravek:
Maybe its more a theme problem?

TerrniTLLC:
I checked in google and firefox as well as on my mobile browser. Everywhere I look there is black text on the element when it is active. It's a small change, but I think it looks better this way

mlmoravek:
Yes, but maybe it should be set as default by the theme?

mlmoravek:
Ahh, I see what the problem is. Thank you @TerrniTLLC :)

PR #864, opened 2024-03-25
fix(autocomplete): update type declaration for selected value in Autocomplete events

andrzejkupczyk:
This fixes a small problem with the autocomplete options/suggestions passed as an array of objects when the `field` prop is not being set. The proposed change is to allow an object as a potential value type. This change aims to prevent linter warnings that occur when binding `update:modelValue` and `select` events:

image

I'm unsure if this is the right approach or not, so I'm open to suggestions.

### DIFF
--- a/packages/docs/components/Autocomplete.md
+++ b/packages/docs/components/Autocomplete.md
@@ -85,20 +85,20 @@ title: Autocomplete
 
 ### Events
 
-| Event name        | Properties                                             | Description                                    |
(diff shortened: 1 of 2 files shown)


mlmoravek:
@andrzejkupczyk Yes, there is a type problem, but I think its bigger than just those two emits.

andrzejkupczyk:
I see the secondary pull request—thank you so much for all those fixes ❤️

PR #866, opened 2024-03-25
fix(autocomplete): change options type definition to generic

mlmoravek:
## Proposed Changes

- add autocomplete options type generics
- fixed "update:modelValue" and "select" (thanks to @andrzejkupczyk)


packages/oruga/src/components/taginput/Taginput.vue
```diff
@@ -223,24 +231,24 @@ const props = defineProps({
 const emits = defineEmits<{
     /**
      * modelValue prop two-way binding
-     * @param value {any[]} updated modelValue prop
+     * @param value {(String | Number | Object)[]} updated modelValue prop
```
andrzejkupczyk, line 234: Just a minor suggestion to use primitive types here, to align with changes within `Autocomplete`, common JS practices and the types expected by TypeScript.

```suggestion
     * @param value {(string | number | object)[]} updated modelValue prop
```

packages/oruga/src/components/taginput/Taginput.vue
```diff
@@ -223,24 +231,24 @@ const props = defineProps({
...
      * on input change event
-     * @param value {any} input value
+     * @param value {String} input value
      */
-    (e: "input", value: any): void;
+    (e: "input", value: string): void;
     /**
      * new item got added
-     * @param value {any} added item
+     * @param value {String | Number | Object} added item
```
andrzejkupczyk, line 244:
```suggestion
     * @param value {string | number | object} added item
```

packages/oruga/src/components/taginput/Taginput.vue
```diff
@@ -223,24 +231,24 @@ const props = defineProps({
...
      */
-    (e: "add", value: any): void;
+    (e: "add", value: T): void;
     /**
      * item got removed
-     * @param value {any} removed item
+     * @param value {String | Number | Object} removed item
```
andrzejkupczyk, line 249:
```suggestion
     * @param value {string | number | object} removed item
```

packages/oruga/src/components/autocomplete/Autocomplete.vue
```diff
@@ -351,14 +351,12 @@ function setItemRef(
 const { checkHtml5Validity, onInvalid, onFocus, onBlur, isFocused, setFocus } =
     useInputHandler(inputRef, emits, props);
 
-const vmodel = useVModelBinding<string | number>(props, emits, {
-    passive: true,
-});
+const vmodel = defineModel<string | number>({ default: undefined });
```
andrzejkupczyk, line 354: I didn't know about generic components, that's great! So, if we make the suggested change, do you think it will work? Right now, I think we're missing a type.

```suggestion
const vmodel = defineModel<T>({ default: undefined });
```

packages/oruga/src/components/autocomplete/Autocomplete.vue
```diff
@@ -56,7 +56,7 @@ const props = defineProps({
     /** Override existing theme classes completely */
     override: { type: Boolean, default: undefined },
     /** @model */
-    modelValue: { type: [String, Number], default: "" },
+    modelValue: { type: [String, Number], default: undefined },
```
andrzejkupczyk, line 59: In TypeScript, it is generally not recommended to declare an object type using `any` for most situations. However, I have noticed this practice in Oruga's codebase multiple times, so it seems acceptable in this case 👍🏻 

```suggestion
    modelValue: { type: [String, Number, Object], default: undefined },
```

andrzejkupczyk (approved):

andrzejkupczyk (commented):
Once again, thank you for those changes—I really appreciate it. However, I think we might be missing the `Object` type in one more place: the formatter definition (`value: string | number`). Unfortunately, I can't add a comment directly to that line since it hasn't been modified recently.

packages/oruga/src/components/autocomplete/Autocomplete.vue
mlmoravek: No, that was a misunderstanding I had at first too. The vmodel here is not the selected option. The vmodel is the input value, which can be an option or a property, defined by the field prop, of an option. So the vmodel is a string because the input value is a string (I think the o-input does not work for number, even a number type is acceptable, but that is a topic for later).

Currently the selected option is only emitted by the `select` event, and doesn't have a two-way binding. 
Maybe we should change that? But this will be a breaking change and will only be introduced when we make the 0.9 release.

andrzejkupczyk: Ah, got it. Thanks for the explanation.

PR #870, opened 2024-03-27
fix(tabs): fix disable animation

### LINKED ISSUE #863: Unable to disable animation on `Tabs` component
### Overview of the problem

**Oruga** version: 0.8.6
**Vuejs** version: 3.4.21
**OS/Browser**: macOS/Chrome

### Description

It appears that it is no longer possible to disable animation after upgrading from Oruga 0.7 to 0.8.6.

### Steps to reproduce

```html
<OTabs v-model="activeTab" :animated="false">
  <OTabItem v-for="tab in tabs" :key="tab.id" :value="tab.id" :label="tab.label" tabindex="-1">
    <Component :is="tab.component" />
  </OTabItem>
<OTabs>
```

### Expected behavior

When tab is clicked, the animation should be disabled.

### Actual behavior

The animation is still enabled.

messenjer:
Fixes #863

## Proposed Changes

Replace `disabled` by `css`, the prop to disable transition is `css` https://vuejs.org/api/built-in-components.html#transition

### DIFF
--- a/packages/oruga/src/components/steps/StepItem.vue
+++ b/packages/oruga/src/components/steps/StepItem.vue
@@ -164,7 +164,7 @@ const elementClasses = defineClasses(["itemClass", "o-steps__item"]);
 
 <template>
     <Transition
-        :disabled="!parent.animated"
+        :css="parent.animated"
         :name="transitionName"
         :appear="parent.animateInitially"
         @after-enter="afterEnter"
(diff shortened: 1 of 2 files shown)


messenjer:
I fixed also the same issue in the StepItem component : https://github.com/oruga-ui/oruga/pull/870/commits/e58670e30038002903cf2f9265d5141d5f38c04a

PR #877, opened 2024-03-29
Bugfix: Remove toRaw from TableColumn data provider when providing props

### LINKED ISSUE #876: Table Column Visibility Not Updating When Visible Is Changed
### Overview of the problem

Toggling visible on table columns does not cause the table column to hide. 

**Oruga** version: 0.8.6
**Vuejs** version: 3.4.21   
**OS/Browser**: Firefox/Chrome

### Description

When the visible property of the table columns is reactive, and changed between true and false the columns do not hide.

Static values of true or false do control the column visibility successfully.  It seems this works in Vue 3.2.41 but does not work in 3.4.21 and may be related to rerender efficiency improvements in Vue. Specifically [this line](https://github.com/oruga-ui/oruga/blob/24f70ce4954253f97d1532371fc7e1fa42fb90c2/packages/oruga/src/components/table/TableColumn.vue#L91) spreads the props using the toRaw function which breaks the reactivity of the props.  Removing the toRaw and just spreading props as-is fixes the bug.

columns

### Steps to reproduce

1. Navigate to Oruga Examples [Table Detailed](https://oruga-ui.com/components/Table.html#detailed).
2. In the detail example table try toggling the "Name" column and observe that the Name column does not hide.  The boardgame names do hide in the detail pane but that pane does not use oruga table components.

### Expected behavior

Table columns to hide when visibility prop is programmatically toggled to false
### Actual behavior

Column visibility always maintains the initial state of the visibility prop.

daleshort:
Fixes oruga-ui/oruga/issues#876

Fixes #876 

## Proposed Changes
Remove toRaw when providing property data to restore reactivity to prop changes.

### DIFF
--- a/packages/oruga/src/components/table/TableColumn.vue
+++ b/packages/oruga/src/components/table/TableColumn.vue
@@ -89,6 +89,7 @@ const vm = getCurrentInstance();
 
 const providedData = computed<TableColumnComponent>(() => ({
     ...toRaw(props),
+    visible: props.visible,
     $el: vm.proxy,
     $slots: vm.slots,
     style: style.value,


mlmoravek:
solved by https://github.com/oruga-ui/oruga/pull/878

PR #882, opened 2024-04-02
fix(datepicker): make date parsing work again

### LINKED ISSUE #881: Datepicker component no longer editable
### Overview of the problem

**Oruga** latest
**Vuejs** version: latest
**OS/Browser**: any

### Description

In the vue 2 version of oruga, the `o-datepicker` component used to have a prop named `editable`, which as its name suggested, allowed for the inline editing of the date. Now, in the latest version of oruga for vue 3, that prop is gone, and the only prop that resembles that is `readonly`, which acts as the native `readonly` attribute.

### Steps to reproduce

I created a [sandbox reproducing the issue](https://codesandbox.io/p/devbox/oruga-vue-3-vite-5-6wcymh).

### Expected behavior

When the date is inline edited, if the text is a valid date, it should be set as the date.

### Actual behavior

The date is either not changed or emptied.

blm768:
Fixes #881

## Proposed Changes

- Accept plain dates in `onChange`, not just date-range pairs

Note that parsing of date ranges is still broken; it collapses them to a single date.
That seems like a separate bug, though.

### DIFF
--- a/packages/oruga/src/components/datepicker/Datepicker.vue
+++ b/packages/oruga/src/components/datepicker/Datepicker.vue
@@ -852,13 +852,15 @@ function formatNative(value: Date | Date[]): string {
 /** Parse string into date */
 function onChange(value: string): void {
     const date = (props.dateParser as any)(value, defaultDateParser);
+    const validDate = (d: unknown): d is Date =>
+        d instanceof Date && !isNaN(d.getTime());
 
     if (
-        date instanceof Date ||
+        validDate(date) ||
         (Array.isArray(date) &&
             date.length === 2 &&
-            date[0] instanceof Date &&
-            date[1] instanceof Date)
+            validDate(date[0]) &&
+            validDate(date[1]))
     ) {
         vmodel.value = date;
     } else {
--- a/packages/oruga/src/components/datepicker/Datepicker.vue
+++ b/packages/oruga/src/components/datepicker/Datepicker.vue
@@ -854,11 +854,11 @@ function onChange(value: string): void {
     const date = (props.dateParser as any)(value, defaultDateParser);
 
     if (
-        date &&
-        Array.isArray(date) &&
-        date.length === 2 &&
-        !isNaN(date[0]) &&
-        !isNaN(date[1])
+        date instanceof Date ||
+        (Array.isArray(date) &&
+            date.length === 2 &&
+            date[0] instanceof Date &&
+            date[1] instanceof Date)
     ) {
         vmodel.value = date;
     } else {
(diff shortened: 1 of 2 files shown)


packages/oruga/src/components/datepicker/Datepicker.vue
```diff
@@ -854,11 +854,11 @@ function onChange(value: string): void {
     const date = (props.dateParser as any)(value, defaultDateParser);
 
     if (
-        date &&
-        Array.isArray(date) &&
-        date.length === 2 &&
-        !isNaN(date[0]) &&
-        !isNaN(date[1])
+        date instanceof Date ||
```
mlmoravek, line 857: Pls change this line to: `(date instanceof Date && !isNaN(date.getTime()))`

packages/oruga/src/components/datepicker/Datepicker.vue
```diff
@@ -854,11 +854,11 @@ function onChange(value: string): void {
     const date = (props.dateParser as any)(value, defaultDateParser);
```
mlmoravek, line 854: Could you here add a specific type definition `const date: Date = ` please :)

mlmoravek (commented):
Maybe we should also change the default value of `readonly` to `false` instead of true? What do you think?

packages/oruga/src/components/datepicker/Datepicker.vue
blm768: At this point, `date` might not be a `Date`, though; it could also be a `[Date, Date]`  (at least if `defaultDateParser` handled ranges correctly) or `null`. It seems better to me to do an explicit type check.

packages/oruga/src/components/datepicker/Datepicker.vue
blm768: Fixed [here](https://github.com/oruga-ui/oruga/pull/882/commits/4c1de19279d59c3a9be7188337454221c38fc803).

packages/oruga/src/components/datepicker/Datepicker.vue
```diff
@@ -852,13 +852,15 @@ function formatNative(value: Date | Date[]): string {
 /** Parse string into date */
 function onChange(value: string): void {
     const date = (props.dateParser as any)(value, defaultDateParser);
+    const validDate = (d: unknown): d is Date =>
```
mlmoravek, line 855: What do you think about moving this to `utils/helper.ts` and name it `isDate` like the existing helper `isObject` ?

mlmoravek:
@blm768 I appreciate that you are adding tests related to the bug! :)

blm768:
> @blm768 I appreciate that you are adding tests related to the bug! :)

You're welcome. It was a good chance for me to learn how to use `@vue/test-utils`; I need to start using that with my own Vue projects.

PR #890, opened 2024-04-10
fix(Menu): ref unwrapping error

### LINKED ISSUE #885: Error undefined value when clicking menu item
### Overview of the problem

**Oruga** version: [0.8.7]
**OS/Browser**: Firefox / Chrome

### Description

When clicking on an o-menu element, js console throw a TypeError: item.data.value is undefined
Trace logs indicates the bug lies in the resetMenu function in Menu.vue

### Steps to reproduce

On oruga-ui.com in the documentation page for menu click on a menu element. 

### Expected behavior

no js error

dauriata:
Fixes #885 

## Proposed Changes

This is a quick fix to make Menu work again but there is a typescript error that I don't understand and I needed to add a ts ignore line.
I think this is a problem with automatic unwrapping of ref. 
I tried to mimic the part of code in a simple vue script and ts got it right and applied the unwrapping of ref, I don't understand what get TypeScript to behaves differently with oruga code.

### DIFF
--- a/packages/oruga/src/components/menu/Menu.vue
+++ b/packages/oruga/src/components/menu/Menu.vue
@@ -95,7 +95,8 @@ const { childItems } = useProviderParent<MenuItemComponent>(rootRef, {
 
 function resetMenu(excludedItems: ProviderItem[] = []): void {
     childItems.value.forEach((item) => {
-        if (!excludedItems.includes(toRaw(item))) item.data.value.reset();
+        // @ts-ignore
+        if (!excludedItems.includes(toRaw(item))) item.data.reset();
     });
 }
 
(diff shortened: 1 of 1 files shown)


dauriata:
Found out that removing this type assertion : 
 `const childItems = ref<ProviderItem<ItemData>[]>([]) as Ref<
        ProviderItem<ItemData>[]
    >;`
 in useParentProvider.ts  prevent the error and give a correct type when accessing inner ItemData , but create ts error when looking at the childtems type.

So instead of ignoring TS check I changed to used type assertion too.

mlmoravek:
@dauriata Yes, you are right. It seems that a ref is unpacked when it is added to another ref.
https://vuejs.org/guide/essentials/reactivity-fundamentals.html#additional-ref-unwrapping-details

So it's a slightly more complex type problem.

PR #912, opened 2024-04-27
fix(table): debounce function not called

dauriata:
## Proposed Changes

Fixes debounce function in table search filter.
The function was not called


mlmoravek:
@dauriata Thanks for fixing all this table problems!

dauriata:
I added a commit making `debounceSearch `a global config option

mlmoravek:
@dauriata Please regenerate the docs page so that the prop documentation can be changed.

PR #919, opened 2024-05-01
feat(table): enhance table with generics

### LINKED ISSUE #701: Table: Use data mapping object instead of merging values
### Overview of the problem

**Oruga** version: [0.8.X]

### Description

In the current implementation of the table component, each element of the table data property (which is a user-defined array) is extended by a special identifier to implement specific behaviour of the table. 

But this is a bad concept.  It also results in unexpected behaviour for some users.

We shouldn't be touching the data props to set or recreate data.
We can resolve it without using an Object.assign to avoid cloning each row.
We can use an object map where the key is the row and the value is uuid.
Or table data is an array of our class/interfaces, 

```
interface TableData { 
  key: any;
  data: any[];
}
```
In the future 
```
TableData<T> {
  id: any;
  data: T[];
} 
```

### LINKED ISSUE #924: Feature(table) : Give access to filtered table length
I would need to display the resulting number of rows of a table after filtering with the search field and without backend filtering.
Migrating from buefy I used to get access directly to child property of the data filtered. With composition API and script setup now all inner properties are private.

 A solution would be to add 
`defineExpose( {tableRows} )` at the end of the script block
or implement an event `update:filteredTableRows` that would return filtered data

Would be glad to contribute with a PR if there is an implementation preference

mlmoravek:
Fixes #701
Fixes #924

## Proposed Changes

- add generic to table
  - update props
  - update emits
- wrap table data into object mapping
- add `processed` emit which emits every time after the rows got processed
- refactor table
- update export types
- update examples


dauriata:
@mlmoravek I just checked out your PR and for your information on the table example page it seems to me that seachable is not working and dragging throws a js error. 

Also I looked at the code and I wonder why is it needed to compute an uuid, why not use the original data array row index as it seems to me that the original data array is never mutated ?

mlmoravek:
@dauriata Thanks for the review! You are right, there is a problem with the search function. 
But dragging works as expected for me. 

I will see if I can use the index instead of the uuid. At first I wrapped each row item completely and just emitted and worked with the wrapper item. But I changed it back. Maybe I can undo the uuid calculation too.

PR #930, opened 2024-05-21
feat(inputs): detect constraint validation attribute changes

blm768:
When attributes related to HTML constraint validation change on an input, they may change or even remove the validation error on the input. Ideally, we'd reflect the change of validation error in the enclosing field.

## Proposed Changes

Changes to the following attributes on an invalid form element should trigger a call to `checkHtml5Validity`:

- `required`
- `pattern`
- `maxlength`
- `minlength`
- `max`
- `min`
- `step`

Additionally, changes to the `disabled` attribute on the input or any ancestor `<fieldset>` should trigger a call to `checkHtml5Validity`.

## Sample

```vue
<script setup lang="ts">
import { OCheckbox, OInput } from '@oruga-ui/oruga-next';
import { ref } from 'vue';

const val = ref("");
const required = ref(true);
const disabled = ref(false);
const groupDisabled = ref(false);
</script>

<template>
  <main>
    <OField label="Value">
      <fieldset :disabled="groupDisabled">
        <OInput v-model="val" :required :disabled />
      </fieldset>
    </OField>
    <OField>
      <OCheckbox v-model="required">Required</OCheckbox>
    </OField>
    <OField>
      <OCheckbox v-model="disabled">Disabled (directly)</OCheckbox>
    </OField>
    <OField>
      <OCheckbox v-model="groupDisabled">Disabled (group)</OCheckbox>
    </OField>
  </main>
</template>
```

### Before

[before.webm](https://github.com/oruga-ui/oruga/assets/2198455/151beef3-12a4-4d47-8a0d-795cda06541a)

### After

[after.webm](https://github.com/oruga-ui/oruga/assets/2198455/d2e78568-34b9-4f12-b645-4f92f22abbbb)

### DIFF
--- a/packages/oruga/src/composables/useInputHandler.ts
+++ b/packages/oruga/src/composables/useInputHandler.ts
@@ -138,7 +138,7 @@ export function useInputHandler(
      * If validation fail, send 'danger' type,
      * and error message to parent if it's a Field.
      */
-    function checkHtml5Validity(): boolean {
+    function checkHtml5Validity(): void {
         if (!props.useHtml5Validation) return;
 
         if (!element.value) return;
@@ -149,8 +149,6 @@ export function useInputHandler(
             setInvalid();
             isValid.value = false;
         }
-
-        return isValid.value;
     }
 
     function setInvalid(): void {
--- a/packages/oruga/src/composables/useInputHandler.ts
+++ b/packages/oruga/src/composables/useInputHandler.ts
@@ -37,6 +37,7 @@ function asValidatableFormElement(el: unknown): ValidatableFormElement | null {
 }
 
 const constraintValidationAttributes = [
+    "disabled",
     "required",
     "pattern",
     "maxlength",
@@ -238,6 +239,7 @@ export function useInputHandler(
             if (!isValid.value) checkHtml5Validity();
         };
         let validationAttributeObserver: MutationObserver | null = null;
+        let ancestorMutationObserver: MutationObserver | null = null;
         watch(
             [maybeElement, isValid, () => props.useHtml5Validation],
             (data) => {
@@ -254,10 +256,17 @@ export function useInputHandler(
                     }
                     validationAttributeObserver.disconnect();
                 }
+                if (ancestorMutationObserver != null) {
+                    if (ancestorMutationObserver.takeRecords().length > 0) {
+                        onAttributeChange();
+                    }
+                    ancestorMutationObserver.disconnect();
+                }
 
                 if (el == undefined || valid || !useValidation) {
                     return;
                 }
+
                 if (validationAttributeObserver == null) {
                     validationAttributeObserver = new MutationObserver(
                         onAttributeChange,
@@ -266,6 +275,25 @@ export function useInputHandler(
                 validationAttributeObserver.observe(el, {
                     attributeFilter: constraintValidationAttributes,
                 });
+
+                // Note that this doesn't react to changes in the list of ancestors.
+                // Based on testing, Vue seems to rarely, if ever, re-parent DOM nodes;
+                // it generally prefers to create new ones under the new parent.
+                // That means this simpler solution is likely good enough for now.
+                let ancestor: Node | null = el;
+                while ((ancestor = ancestor.parentNode)) {
+                    // Form controls can be disabled by their ancestor fieldsets.
+                    if (ancestor instanceof HTMLFieldSetElement) {
+                        if (ancestorMutationObserver == null) {
+                            ancestorMutationObserver = new MutationObserver(
+                                onAttributeChange,
+                            );
+                        }
+                        ancestorMutationObserver.observe(ancestor, {
+                            attributeFilter: ["disabled"],
+                        });
+                    }
+                }
             },
             { immediate: true },
         );
--- a/packages/oruga/src/composables/useInputHandler.ts
+++ b/packages/oruga/src/composables/useInputHandler.ts
@@ -74,6 +74,9 @@ export function useInputHandler(
     // inject parent field component if used inside one
     const { parentField } = injectField();
 
+    /// Allows access to the native element in cases where it might be missing,
+    /// e.g. because the component hasn't been mounted yet or has been suspended
+    /// by a <KeepAlive>
     const maybeElement = computed<ValidatableFormElement | undefined>(() => {
         const el = unrefElement<Component | HTMLElement>(inputRef);
         if (!el) {
@@ -95,6 +98,8 @@ export function useInputHandler(
         return inputs as ValidatableFormElement;
     });
 
+    /// Should be used for most accesses to the native element; we generally
+    /// expect it to be present, especially in event handlers.
     const element = computed(() => {
         const el = maybeElement.value;
         if (!el) {
--- a/packages/oruga/src/composables/useInputHandler.ts
+++ b/packages/oruga/src/composables/useInputHandler.ts
@@ -11,6 +11,7 @@ import { injectField } from "@/components/field/fieldInjection";
 import { unrefElement } from "./unrefElement";
 import { getOption } from "@/utils/config";
 import { isSSR } from "@/utils/ssr";
+import { isDefined } from "@/utils/helpers";
 
 // This should cover all types of HTML elements that have properties related to
 // HTML constraint validation, e.g. .form and .validity.
@@ -263,7 +264,7 @@ export function useInputHandler(
                     ancestorMutationObserver.disconnect();
                 }
 
-                if (el == undefined || valid || !useValidation) {
+                if (!isDefined(el) || valid || !useValidation) {
                     return;
                 }
 
--- a/packages/oruga/src/composables/useInputHandler.ts
+++ b/packages/oruga/src/composables/useInputHandler.ts
@@ -240,7 +240,6 @@ export function useInputHandler(
             if (!isValid.value) checkHtml5Validity();
         };
         let validationAttributeObserver: MutationObserver | null = null;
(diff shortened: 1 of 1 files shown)


blm768:
Without in-browser testing of some kind, it's hard to really automate testing of this change. Are there any plans to restore Cypress (or use another in-browser test framework) that require a full browser implementation of the DOM?

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -203,6 +220,85 @@ export function useInputHandler(
...
+                    validationAttributeObserver.disconnect();
+                }
+                if (ancestorMutationObserver != null) {
+                    if (ancestorMutationObserver.takeRecords().length > 0) {
+                        onAttributeChange();
+                    }
+                    ancestorMutationObserver.disconnect();
+                }
+
+                if (el == undefined || valid || !useValidation) {
```
mlmoravek, line 266: you can use the helper `isDefined` for `el == undefined` if you like :)

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -203,6 +220,85 @@ export function useInputHandler(
...
+                // Note that this doesn't react to changes in the list of ancestors.
+                // Based on testing, Vue seems to rarely, if ever, re-parent DOM nodes;
+                // it generally prefers to create new ones under the new parent.
+                // That means this simpler solution is likely good enough for now.
+                let ancestor: Node | null = el;
+                while ((ancestor = ancestor.parentNode)) {
+                    // Form controls can be disabled by their ancestor fieldsets.
+                    if (ancestor instanceof HTMLFieldSetElement) {
+                        if (ancestorMutationObserver == null) {
+                            ancestorMutationObserver = new MutationObserver(
```
mlmoravek, line 288: I haven't worked much with observers yet. Is it really necessary to have multiple observers, or is it possible to have one observer for multiple targets?

mlmoravek:
@blm768 I really appreciate your clear and well written PRs and Issues. You have some very nice requests for improvements for Oruga!

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -35,6 +36,17 @@ function asValidatableFormElement(el: unknown): ValidatableFormElement | null {
...
+const constraintValidationAttributes = [
+    "disabled",
+    "required",
+    "pattern",
+    "maxlength",
+    "minlength",
+    "max",
+    "min",
+    "step",
+];
```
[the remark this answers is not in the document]
blm768: I can do that, but [this line](https://github.com/oruga-ui/oruga/pull/930/files/126763ed1fb2e3d96c0605ea53b54a68b9993d0c#diff-e8e2be411f6a4c43f5676dd3bffe86bda2ad8c1420749d6376f37f7b8a75566dR276) then starts complaining because the `as const` implicitly makes the array `readonly`. TypeScript's built-in DOM types don't mark `attributeFilter` as `readonly`, so the object construction is marked as an error. Based on the DOM API specification, the parameter is a WebIDL `sequence`, which is passed by copy and not mutated in-place, so TypeScript _could_ safely mark that part of the call signature as `readonly`, but it hasn't done so yet.

packages/oruga/src/composables/useInputHandler.ts
blm768: One observer can be attached to multiple targets; in fact, `ancestorMutationObserver` is tied to several DOM elements (every ancestor of the form control). Now that you mention it, though, I do think we can use the same observer to watch both the form element (and all its relevant attributes) and the ancestors (for just the `disabled` attribute). The two-observer strategy is partially a hold-over from a much more elaborate solution I was working on, which tried to detect cases where the form control got re-parented into another part of the DOM. Since Vue doesn't seem to re-parent nodes in typical use cases, I abandoned that approach to avoid the complexity.

I've pushed up [a patch](https://github.com/oruga-ui/oruga/pull/930/commits/dcd62557f5fa80eb36c7b862e676771278aeeef8).

blm768:
> @blm768 I really appreciate your clear and well written PRs and Issues. You have some very nice requests for improvements for Oruga!

Thanks! Oruga has been providing significant value for me and the rest of the team I work with, so it's only fair that I contribute back, especially when it makes Oruga more suitable for my own use cases.

mlmoravek:
> Without in-browser testing of some kind, it's hard to really automate testing of this change. Are there any plans to restore Cypress (or use another in-browser test framework) that require a full browser implementation of the DOM?

At the moment there are no plans to restore Cypress until most of the components have been unit tested. I think we have enough open stuff to do with the current stack until we can look further into more testing techniques.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -149,8 +168,6 @@ export function useInputHandler(
             setInvalid();
             isValid.value = false;
         }
-
-        return isValid.value;
```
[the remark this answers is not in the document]
mlmoravek: Ohh yeah, true!

mlmoravek:
> Thanks! Oruga has been providing significant value for me and the rest of the team I work with, so it's only fair that I contribute back, especially when it makes Oruga more suitable for my own use cases.

This is how I became a maintainer for oruga :D

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -61,10 +73,9 @@ export function useInputHandler(
     // inject parent field component if used inside one
     const { parentField } = injectField();
 
-    const element = computed<ValidatableFormElement>(() => {
+    const maybeElement = computed<ValidatableFormElement | undefined>(() => {
         const el = unrefElement<Component | HTMLElement>(inputRef);
         if (!el) {
-            console.warn("useInputHandler: inputRef contains no element");
```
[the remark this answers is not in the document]
mlmoravek: Because you run the watcher on immediate `{ immediate: true },`. Maybe this isn't necessary? While not having immediate, isn't the watcher called after mount, when the element ref is filled?

Otherwise, please add a comment to `maybeElement` about why it exists.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -203,6 +221,73 @@ export function useInputHandler(
...
+        // Technically, having the `required` attribute on one element in a radio button
+        // group affects the validity of the entire group.
+        // See https://html.spec.whatwg.org/multipage/input.html#radio-button-group.
+        // We're not checking for that here because it would require more expensive logic.
+        // Because of that, this will only work properly if the `required` attributes of all radio
+        // buttons in the group are synchronized with each other, which is likely anyway.
+        // (We're also expecting the use of radio buttons with our default validation message handling
+        // to be fairly uncommon because the overall visual experience is clunky with such a configuration.)
+        const onAttributeChange = (): void => {
+            if (!isValid.value) checkHtml5Validity();
```
mlmoravek, line 240: So this is called every time a "constraintValidationAttributes" is changed on the input or its ancestor, right? 
Then I think the "if" check for invalid is wrong. 
Think about having required `false` so the input is valid; then changing required to `true` and the input stays valid, but it shouldn't, because it's required now. Am I right or am I missing something?

blm768:
> At the moment there are no plans to restore Cypress until most of the components have been unit tested. I think we have enough open stuff to do with the current stack until we can look further into more testing techniques.

That's fair enough; there's certainly plenty of room for expanding the current tests.

packages/oruga/src/composables/useInputHandler.ts
blm768: Yes, it's called on every change of one of those attributes. My observation of Oruga's general behavior around validation is that if a change forces an invalid input to become valid, the validation message will be immediately cleared, signaling to the user that the issue has been corrected. On the other hand, if a change forces a previously _valid_ input to become _invalid_, displaying the validation message is deferred until the next "interaction" with the control — where "interaction" is defined as blurring the control or submitting its containing form. Checks [like this one](https://github.com/oruga-ui/oruga/blob/6f1a8b31fe1bebc9a40170320d1250c75046fb42/packages/oruga/src/components/input/Input.vue#L293) are scattered throughout the codebase to achieve that behavior.

This is largely consistent with the native behavior of constraint validation except that purely native constraint validation doesn't react to blur events; it only reacts to form submissions.

packages/oruga/src/composables/useInputHandler.ts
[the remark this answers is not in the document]
blm768: > While not having immediate, isn't the watcher called after mount, when the element ref is filled?

That's the most likely trigger for the watcher, but since it watches multiple values, it may still be triggered before the component is mounted. In particular, changes to `props.useHtml5Validation` might theoretically trigger the watcher prematurely.

By calling the [`onMounted` hook](https://vuejs.org/api/composition-api-lifecycle.html#onmounted), I theoretically could defer the creation of the watcher until the element ref has been populated, but if the component is inside a `<KeepAlive>`, the element ref may later become empty because the component [is detached from the DOM](https://vuejs.org/guide/built-ins/keep-alive.html#lifecycle-of-cached-instance).

I've [added a comment](https://github.com/oruga-ui/oruga/pull/930/commits/e17a4fa0581c561a72441ec4ddf3178a4c8036ea) detailing when we expect the ref to be empty.

packages/oruga/src/composables/useInputHandler.ts
mlmoravek: Thanks for making that clear. You seem to be right.

mlmoravek (approved):

PR #932, opened 2024-05-28
fix(datetimpicker): fix inline value edit | make picker props reactive

### LINKED ISSUE #922: datetimepicker: Editing directly the datetime value does nothing
### Overview of the problem

**Oruga** version: 0.8.9
**Vuejs** version: 3.4.20
**OS/Browser**: Tried on Chromium 126 and Firefox 125

### Description

If a user edit manually the date in the `<input type='text' [...]>` component of the datetimepicker (so without using the calendar component), the new date is well displayed but completely ignored.

### Steps to reproduce

- Load https://oruga-ui.com/components/Datetimepicker.html
- In one of the example select a date with the calendar
- Edit the text value and change the date

image

image

More info here : https://framagit.org/framasoft/mobilizon/-/issues/1443

### Expected behavior

Best :
- the new written date is parsed and the calendar component is updated
- if the date is wrong the user is informed of the problem

At least : 
- it is impossible to manually edit the displayed date by making the `<input type='text' [...]>`  component disabled
- in this case, I suggest to don't use a  `<input>` but a text because there is nothing to input here.
 
### Actual behavior

The user think the date is edited and it is not the case, so the recorded date is not the date he see.

mlmoravek:
Fixes #922

## Proposed Changes

- Remove `this` leftover from Option API to Composition API refactoring
- Enhance timepicker `hourFormat` prop type
- Fix datetimepicker pm/am reference by making timepicker and datepicker props reactive


Lahaxa:
Thanks for the fix!  I read it's fixed but not released, any idea when we it will be released? https://framagit.org/framasoft/mobilizon/-/issues/1443

mlmoravek:
@Lahaxa I will check if we have any other potential issues and PR to include in the release and will make the release soon!

PR #964, opened 2024-06-17
fix(select): adjust placeholder behavior

### LINKED ISSUE #921: HTML5 required validity does not work with OSelect component with a placeholder
### Overview of the problem

**Oruga** version: [0.7.0]
**Vuejs** version: [3.3.13]
**OS/Browser**: Firefox 125.0.3 on macOS

### Description

When using an OSelect component with a placeholder, the component shows as "valid" when no option has been chosen.

### Steps to reproduce
```
const value = ref(null);
const mySelectRef = ref(null);

<OSelect ref="mySelectRef" v-model="value" placeholder="Select me!">
  <option value="pickme">Pick me</option>
</OSelect>
```

### Expected behavior

mysSelectRef.checkHtml5Validity() === false

### Actual behavior

mysSelectRef.checkHtml5Validity() === true

### Additional details

Upon inspecting things further the "value" of the select element is the placeholder text "Select me!" instead of being blank. This is due to the option having the :value="null" thus no value attribute is actually set on the option so the browser falls back to the inner text of the option.

This also occurs in the current 0.8.9 and also occurs in Google Chrome. Thus it is not isolated to a specific oruga version nor browser.

### Possible solutions

1) use empty string "" as the not-selected value.  This will fix the issue in part as the browser will correctly understand the chosen value. But it messes with the modelValue if the user really wants null to be unselected.
2) customize checkHtml5Validity() for this component handle this situation explicitly.

mlmoravek:
Fixes #921

## Proposed Changes

- cast vmodel property to empty string if its `null` or `undefined`
- use empty string as placebholder value


packages/oruga/src/components/select/examples/base.vue
```diff
@@ -13,56 +20,18 @@ const selectedOptions = ref([]);
...
-                <option value="silver">Silver</option>
-            </o-select>
-        </o-field>
-
-        <o-field label="Rounded">
-            <o-select placeholder="Select a character" rounded>
-                <option value="flint">Flint</option>
-                <option value="silver">Silver</option>
-            </o-select>
+            <o-select placeholder="Select a character" :options="options" />
```
blm768: Moving forward, is the `options` prop the recommended way to specify options for an `OSelect`? I'm not particularly opposed to that, as it provides a cleaner way to map between the "logical" property values and the strings used as the actual option values in the DOM. It does feel a little odd to have two different ways to specify the options, each of which has different functionality available (mapping between strings and arbitrary values for the `options` prop, grouping and more direct DOM control for `<option>` elements); perhaps the long-term goal should be to unify those somehow.

packages/oruga/src/components/select/Select.vue
```diff
@@ -230,7 +230,7 @@ const { parentField, statusVariant, statusVariantIcon } = injectField();
 
 const vmodel = defineModel<T | T[]>({ default: undefined });
 
-const placeholderVisible = computed(() => vmodel.value === null);
+const placeholderVisible = computed(() => !vmodel.value);
```
blm768: I think this may technically be a change in behavior if the `v-model` value is zero.

packages/oruga/src/components/select/Select.vue
```diff
@@ -222,31 +222,41 @@ const emits = defineEmits<{
...
 // inject parent field component if used inside one
 const { parentField, statusVariant, statusVariantIcon } = injectField();
 
-const vmodel = defineModel<T | T[]>({ default: undefined });
+// if id is given set as `for` property on o-field wrapper
+if (props.id) parentField?.value?.setInputId(props.id);
 
-const placeholderVisible = computed(() => !vmodel.value);
+const vmodel = defineModel<T | T[]>({
+    get: (v) => (isDefined(v) ? v : props.multiple ? [] : ""),
```
blm768, line 235: I assume this affects the value emitted with the `update:modelValue` event, which might break some code that was expecting to get `null` values out. In particular, if the user is trying to use objects as the `modelValue`, they'd probably be quite surprised to get an empty string instead of `null` when the placeholder is selected. I'd say that's more a concern when dealing with the `options` prop (which provides a way to use object-valued options in a way that's actually compliant with Web standards) than it would be when using `<option>` elements (which are supposed to be strictly string-valued).

blm768 (changes requested):

packages/oruga/src/components/select/examples/base.vue
mlmoravek: Personally, I like to leave the rendering part to the component and just provide the data. So I added the `options` prop. However, there is no preferred way, either using the `options` prop or passing HMLT option elments inside the default slot.
You can pass any value when using the `options` prop, just like when using the HTML option.

It's kinda like other components; like the Table component for example, where you can provide the TableColumn components with the default slot or pass them as objects via the Table `data` prop.

packages/oruga/src/components/select/Select.vue
mlmoravek: Or you are right! Thank you! I will change this to be more explicit.

packages/oruga/src/components/select/Select.vue
mlmoravek: Did I get the 'get' and 'set' wrong?
I wanted to set the internal representation of vmodel to `""` when it is `null` or `undefined` and pass `null` out of the select component when it is `""`.

blm768: Based on the documentation, I'd expect `get` convert from the internal representation and `set` to convert _to_ the internal representation, so setting `v-model` to `""` would make the internal representation `null`, but the value exposed to events would be `""`. It sounds like you want the reverse, although that would break clients that do want set `v-model` to the empty string and expect to get an empty string back, which matches the standard behavior for a `<select>`.

I haven't tested any of this, though, so I could be wrong about the actual behavior.

packages/oruga/src/components/select/Select.vue
```diff
@@ -237,7 +237,9 @@ const vmodel = defineModel<T | T[]>({
     default: null,
 });
 
-const placeholderVisible = computed(() => !props.multiple && !vmodel.value);
+const placeholderVisible = computed(
+    () => !props.multiple && !isDefined(vmodel.value),
```
blm768, line 241: Do we want to display the placeholder if `vmodel.value` is the empty string? (I guess it may depend on how the `v-model` mapping ends up looking.)

packages/oruga/src/components/select/Select.vue
blm768: For instances that use the `options` prop, it might make sense to define the mapping based on that rather than forcing a mapping between `null` and strings here.

packages/oruga/src/components/select/examples/base.vue
mlmoravek: Vue extends the value type of select options to also support `objects` (https://vuejs.org/guide/essentials/forms.html#select-options). So, in my opinion, there's not that much difference between defining the options as a prop or as html elements.

packages/oruga/src/components/select/Select.vue
mlmoravek: I think I implemented it correctly. It is the same as the 'computed' implementation. 
I tested it and it worked.

packages/oruga/src/components/select/examples/base.vue
blm768: I kind of think Vue's choice there is a bit of a mistake, but I guess it does sort of work that way. The actual `<option>` values in the DOM are pretty weird (all set to the string `"[object Object]"`, at least in Firefox), which I don't really care for, but Vue does seem to patch things up so it still works.

Now that I look more closely, `OptionItem` doesn't have a separate string value for the `<option>` element, so my comment about that mapping is probably irrelevant.

packages/oruga/src/components/select/Select.vue
blm768: This does seem to work in practice, presumably because the "external" value of `v-model` is what gets passed to the `<select>`.

blm768: As for breaking changes, because of how the placeholder actually works (not allowing selection of the placeholder after a "real" option has been selected), I haven't found any significant breaking changes in what I'd consider "reasonable" use. `<option :value="null">` is kind of a weird case; selecting the null-valued option actually causes the `<OSelect>` to emit its model value as `null`, which surprises me given how `vmodel` has been defined, but trying to set the model value to `null` from the parent component doesn't properly select the null-valued option. Using `<option :value="null">` is probably undesirable anyway, as it has the wrong semantics for constraint validation, so maybe that's not a major concern. The behavior does change for this example, though:

```vue
<script setup lang="ts">
import { OButton, OSelect } from '@oruga-ui/oruga-next';
import { ref, watch } from 'vue';

const val = ref<string | null>(null);

watch(val, v => console.debug(v));

</script>

<template>
  <main>
    <OField label="Value">
      <OSelect v-model="val">
        <option :value="null">null</option>
        <option value="1">1</option>
      </OSelect>
      <OButton :click="val = null">Set null</OButton>
    </OField>
  </main>
</template>
```

PR #1023, opened 2024-07-17
Allow custom constraint validation errors

### LINKED ISSUE #928: Support custom constraint validation (setCustomValidity)
Currently, Oruga provides mechanisms to enable HTML constraint validation, and it even provides a way to override the browser's validation message when the input is invalid (via the `validationMessage` prop). However, this validation message only appears if the element fails a native constraint validation check, and it only applies to the messages that Oruga shows; the element's own constraint validation state is unaffected.

It would be convenient to have an officially supported way to call `setCustomValidity` on Oruga's form control elements. This might either replace or supplement the current `validationMessage` prop. Just passing a string would likely be sufficient, although I've run into race conditions when trying to update a string prop right before submitting a form, so for that use case, we might need some mechanism to either make the call to `setCustomValidity` happen faster (e.g. by passing a validation function and attaching a `flush: "sync"` watcher) or provide an event for which we can wait so we know that constraint validation is in the expected state.

We'd probably also need to hook into the logic that propagates validation messages up to the parent field; any changes in the validation message probably should trigger a call to `checkHtml5Validity` (perhaps excepting cases where the field is currently focused and was previously valid, as it's less annoying to wait for the user to complete their input before displaying a message).

blm768:
Fixes #928

- Inputs have a new `customValidity` prop, which allows setting custom constraint validation errors. Unlike the `validationMessage` prop, this new prop actually marks the element as invalid, preventing form submission.


blm768:
It might be feasible to merge the `customValidity` and `validationMessage` props, as their behavior somewhat overlaps. That would be a breaking change, though, so it would require some further discussion.

blm768:
I'll need to extend this to cover other input types if/when we determine whether this is the right API.

mlmoravek:
Hmm, OK, I get it now. 

But maybe our `customValidity` should be a function that takes the `validityState` (input.validity) as an argument and has to return the validation message as string. Then this function, if given, will be called when the validity check returns false. This allows the user to create different validation messages based on the `validityState`.

How about this?

blm768:
> But maybe our `customValidity` should be a function that takes the [`validityState`](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) (input.validity) as an argument and has to return the validation message as string. Then this function, if given, will be called when the validity check returns false. This allows the user to create different validation messages based on the `validityState`.

That could work, but it would be more complicated than strictly necessary for my main use case. The main goal here isn't to override the validation message for inputs that already fail a built-in validation check (e.g. `required`); the existing `validationMessage` prop generally does that well enough. The goal is to add checks that can't be expressed using the normal validation attributes. For instance:

```vue
<script setup lang="ts">
import { OInput } from '@oruga-ui/oruga-next';
import { computed, ref } from 'vue';

const val = ref<string>('');
const validation = computed(() => {
  const value = val.value;
  const valid = value === '' || Number.parseInt(value) % 2 == 0;
  return valid ? '' : 'Enter a number divisible by two';
});

const onSubmit = (e: Event) => { alert('Submitted'); e.preventDefault(); };

</script>

<template>
  <form @submit="onSubmit">
    <OField label="Value">
      <OInput v-model="val" type="number" required :customValidity="validation" />
    </OField>
  </form>
</template>
```

In this example, any number that isn't a multiple of two will fail the validation checks, preventing form submission.

Your suggested API does seem like it would be an improvement for the existing `validationMessage` prop, and it might make the distinction between the two use cases clearer. I have to wonder if we should just merge the props, though, since the new `customValidity` prop can also override the validation message in cases where some other constraint validation check is failing.

Perhaps the right solution (if we're OK with a breaking change) is to merge the props and allow the merged prop to take any one of the following:

1. A string.
2. A function taking no arguments and returning a string (or `undefined`).
3. A function taking a `ValidityState` and returning a string (or `undefined`).

If we want to minimize breakage, we could make it so the prop doesn't set a custom validation error (via `setCustomValidity`) for otherwise valid inputs when it's given just a string, which would make its behavior for that specific case pretty much match the existing behavior. In the long run, though, I think it's cleanest to call `setCustomValidity` even if there's not a built-in validation error.

mlmoravek:
I am just trying to stay open to a wider range of usecases where possible. 

With the current `validationMessage` prop, you can only set the error message, but you cannot set different error messages for different validation errors.

We are currently on a breaking change release with 0.9.0, so it is fine to merge both features into one new prop and implement it as a breaking change.

So the type would be:
`customValidity?: string | (validityState: ValidityState) => string | undefined;` ?

Yes, we should call `setCustomValidity` when the prop is given and an error occurs. 
Do we need to clean it up when the input is valid again? Like `setCustomValidity("")` after calling it once with a message?

blm768:
> Do we need to clean it up when the input is valid again? Like `setCustomValidity("")` after calling it once with a message?

Yes; without cleaning up the custom message, the control will still think it's in an invalid state.

I've pushed up a new version that supports the function-based syntax.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -286,6 +329,12 @@ export function useInputHandler<T extends ValidatableFormElement>(
             },
             { immediate: true },
         );
+        // The custom validation message may depend on the element's validity
+        // state, which may depend on the element's current value.
+        watch(
```
mlmoravek, line 334: I' not sure if this watcher is really necessary, because each component implements it's own way to check the validation state by calling `checkHtml5Validity`. 

However, if necessary, you will need to extend the `props` prop type definition of this `useInputHandler` composable.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -240,26 +280,29 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
-                    if (validationAttributeObserver.takeRecords().length > 0)
-                        onAttributeChange();
-                    validationAttributeObserver.disconnect();
-                }
+                const useValidation = data[1];
 
-                if (!isDefined(el) || valid || !useValidation) return;
+                if (!isDefined(el) || !useValidation) {
+                    // Clean up previous state.
+                    if (validationAttributeObserver != null) {
```
mlmoravek, line 296: Why move the cleanup inside this `if`? 
I don't know exactly how the browser handles old objects, but now every time we create a `new MutationObserver` we override the old object and don't disconnect the old element. This could be a memmory leak problem until the old instance is cleaned up by the browser, which is out of our hands.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -225,6 +226,45 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
+            const validity = props.customValidity ?? "";
+            if (typeof validity === "string") {
+                element.setCustomValidity(validity);
+            } else {
+                element.setCustomValidity(validity(element.validity));
+            }
+        };
+        // Note that using watchEffect will implicitly pick up any reactive dependencies used
+        // inside props.customValidity, which should help the computed message stay up to date.
+        watchEffect(updateCustomValidationMessage);
```
mlmoravek, line 247: Have you tested if the props are really a reactive object with can watchEffect look at? I'm not sure about that, but i could be.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -225,6 +226,45 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
+
+        // Clean up validation state if we stop controlling it.
+        watch(
+            [maybeElement, (): boolean => props.useHtml5Validation ?? true],
```
mlmoravek, line 251: Do you have a reason to have two watchers watching the same objects? Or just cleaner code? Maybe performance wise it is better to merge them together?

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -240,26 +280,29 @@ export function useInputHandler<T extends ValidatableFormElement>(
         // (We're also expecting the use of radio buttons with our default validation message handling
         // to be fairly uncommon because the overall visual experience is clunky with such a configuration.)
         const onAttributeChange = (): void => {
+            updateCustomValidationMessage();
```
mlmoravek, line 283: Would it be nicer to call this within `checkHtml5Validity`? Then it will also be called when the component triggers the call, like when the input value changes.

packages/oruga/src/composables/useInputHandler.ts
blm768: I hadn't noticed that they were watching the same objects, but I think it's cleaner to have separate watchers; they have mostly independent purposes, and mingling their logic would make the second watcher even harder to read. The second watcher also has different timing (`{ immediate: true }`).

Especially with how rarely these watchers are actually triggered, I expect the performance differences to be negligible.

packages/oruga/src/composables/useInputHandler.ts
blm768: Based on the call sites I've looked at, components that use the composable seem to pass their own `props` objects directly through, so the object itself will be reactive. More importantly, any reactive dependencies inside a `watchEffect` function [are automatically registered](https://vuejs.org/api/reactivity-core.html#watcheffect), so if `props.customValidity` has an internal dependency on a reactive value (e.g. it reads from a `ref`), that will get registered against the watcher as well. This seems to work in practice — with the caveat that if `props.customValidity` has any conditional dependencies (e.g. it only reads a `ref` if another condition is true), those won't get registered unless they're observed _within_ the `watchEffect` context.

packages/oruga/src/composables/useInputHandler.ts
blm768: We probably could call it there. My initial thought was that `checkHtml5Validity` wouldn't be called often enough for all cases. If we're just overriding an existing constraint validation message, it's fine to wait until right before we'd display the message (especially since components in a known invalid state should be calling `checkHtml5Validity` on every change). For the case where we're adding a custom constraint validation message to an otherwise valid input, though, we won't call `checkHtml5Validity` until the input receives a `blur` or `invalid` event, which may be too late if the user submits the form using the Enter key.

We might be able to rely on the call to `watchEffect(updateCustomValidationMessage)`, but as noted above, we'd have to ensure that any new reactive effects that show up with a new call to `updateCustomValidationMessage` are observed within the `watchEffect` context. The only way I've found to do that is to trigger the watcher to run its effect again.

packages/oruga/src/composables/useInputHandler.ts
blm768: Are you talking about the definition [here](https://github.com/oruga-ui/oruga/pull/1023/commits/43efa277be65cbe2d39b47f34ed69c2fa35643c0#diff-e8e2be411f6a4c43f5676dd3bffe86bda2ad8c1420749d6376f37f7b8a75566dR70)?

packages/oruga/src/composables/useInputHandler.ts
blm768: The idea here was that we now need to run the observer even when the input is valid. If we always update the validation message in `checkHtml5Validity`, though, that may no longer be necessary.

We only construct a new `MutationObserver` if `validationAttributeObserver` is `null`, which only happens when the variable is first initialized. We do disconnect and reconnect the observer, but that should actually work as expected in most cases. The first connection [will automatically disconnect all existing connections if the element hasn't changed](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#reusing_mutationobservers). If the element has changed, though, this code would have issues.

The next version I'm working on probably won't need this change anyway, but if it does, there's a little rework I'll want to do here.

packages/oruga/src/composables/useInputHandler.ts
mlmoravek: Ohh yeah, i haven't seen it, sorry.

blm768: It looks like it's _almost_ possible to make all of this work without the watcher on `modelValue`, but the following sample app has issues:

```vue
<script setup lang="ts">
import { ref } from 'vue';

const val = ref<string>('');
function validation(state: ValidityState): string {
  if (state.valueMissing) { // Branch A
    return ''; // Fall back on the browser's default message in this case.
  } else { // Branch B
    return val.value.startsWith('a') ? '' : 'Enter a value that starts with "a".';
  }
}

const onSubmit = (e: Event) => { alert('Submitted'); e.preventDefault(); };
</script>

<template>
  <form @submit="onSubmit">
    <OField label="Value">
      <OInput v-model="val" required :customValidity="validation" />
    </OField>
    <OButton native-type="submit">Submit</OButton>
  </form>
</template>
```

If you enter a value that starts with the wrong character and then immediately hit Enter, the form submission goes through. Normally, the `watchEffect` call would pick up any reactive dependencies, which would include `val`. However, because the input _starts_ in an invalid state, `watchEffect` only sees Branch A the first time it runs the watch function, so it doesn't recognize the dependency on `val`. When you enter a value in the field, the element's validity state immediately changes, but because that's not a reactive value, `watchEffect` doesn't know to run the watcher again, which would run through branch B and start tracking the reactive dependency.

mlmoravek: Hmm maybe we can add the value to the interface of our `customValidity` prop as second or first parameter, then when the value got changed the `watchEffect` will be triggert, right?

blm768: If we retrieve `props.modelValue` in the `watchEffect` callback, it should have the same effect as the watcher regardless of whether we pass it to `props.customValidity` or not, so that could be a viable option. The semantics would be almost exactly the same as having the watcher, but it might slightly reduce the number of duplicate calls to `updateCustomValidationMessage`. I'm slightly leaning toward just retrieving the value for the reactive dependency subscription and throwing it away because then we don't have to directly deal with defining the argument types of `props.customValidity` based on the component's `modelValue` type. How does that sound to you?

mlmoravek: But maybe a user would like to include the value in a custom validation message, therefore i would say adding the modelValue as attribute to the `customValidation` function might be a good solution.

blm768: Maybe, but in basically all cases, I expect they'd have access to the model value anyway, e.g. through the `ref` they're using to drive it. The example I posted earlier is a good example; it accesses the model value without ever needing to have it passed in.

packages/oruga/src/composables/useInputHandler.ts
blm768: Unfortunately, it does look like there are cases where this watcher has to get involved for valid inputs. I've cleaned this section up a bit, though.

blm768:
The diff is (unfortunately) a bit of a mess with the restructuring, but I've added some changes that should make this more robust. This seems to hold up in the test cases I've been able to think of so far.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -224,13 +229,81 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
+     * properties. The `watchEffect` call on `updateCustomValidationMessage` needs to know
+     * about those new reactive dependencies, and the only way to inform it of them is to
+     * re-run `updateCustomValidationMessage` within the watcher's context.
+     *
+     * This does mean that if you need to update the DOM element's validity state _immediately_,
+     * you'll need to call updateCustomValidationMessage directly (to make the change happen now
+     * rather than waiting for the reactivity system) _and_ trigger this `ref` (to register
+     * reactive dependencies), so `updateCustomValidationMessage` will run twice for those updates.
+     */
+    const forceValidationReactivityUpdate = ref(null);
```
mlmoravek, line 250: I don't think we need that here. You are thinking of something too complicated. 

The `watchEffect` calls a function `updateCustomValidationMessage`. 
What you are trying to do here is force the watcher to call this function via a trigger (`forceValidationReactivityUpdate`), then you want to call the trigger with `triggerRef(forceValidationReactivityUpdate);` by yourself.

But why doesn't you just call the function itself? It shouldn't make any difference whether the function is called by the watcher whenever a reactive dependency has changed, or by calling it by hand.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -224,13 +229,81 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
+            // which isn't a reactive property. `element.validity` depends on
+            // the element's current value and the native constraint validation
+            // attributes. We can use `props.modelValue` as a reasonable proxy
+            // for the DOM element's value, and `props.modelValue` _is_ reactive,
+            // so we can use it to help solve that reactivity problem.
+            // When `props.modelValue` is retrieved in this function, even if we
+            // throw the value away, the `watchEffect` call on this function will
+            // register it as a dependency.
+            props.modelValue;
+            element.setCustomValidity(validity(element.validity));
```
mlmoravek, line 278: If you call `props.modelValue` here in any way, just pass it as the second parameter to the `validity` function. That way the user can use it, or just use the first parameter without the second.

packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -240,51 +313,83 @@ export function useInputHandler<T extends ValidatableFormElement>(
         // (We're also expecting the use of radio buttons with our default validation message handling
         // to be fairly uncommon because the overall visual experience is clunky with such a configuration.)
         const onAttributeChange = (): void => {
-            if (!isValid.value) checkHtml5Validity();
+            if (!isValid.value) {
+                // This should update the custom validation message.
+                checkHtml5Validity();
+            } else {
+                triggerRef(forceValidationReactivityUpdate);
```
mlmoravek, line 320: Just call `updateCustomValidationMessage()` here.

mlmoravek:
> This seems to hold up in the test cases I've been able to think of so far.

Maybe you can add some unit test for this composable?

blm768:
> Maybe you can add some unit test for this composable?

It definitely needs some tests, but implementing them outside of a browser context may be tricky. I don't think the current DOM polyfill supports the constraint validation API.

blm768:
Should we consider just adding browser-based testing back in for these kinds of cases? This definitely seems like it's getting complicated enough that it needs browser-level tests.

packages/oruga/src/composables/useInputHandler.ts
blm768: > But why doesn't you just call the function itself? It shouldn't make any difference whether the function is called by the watcher whenever a reactive dependency has changed, or by calling it by hand.

The difference is that Vue only detects new reactive dependencies (e.g. ones that are in a branch that hadn't been called before) if the function is called inside the watcher's context. Before we added a reactive dependency on `props.modeValue`, this code had all sorts of reactivity issues where changes to an unrecognized reactive dependency would make the message stale. Adding that dependency has helped dramatically because changes in the control's value are _usually_ the main thing affecting `element.validityState`, but if (1) there's a branch where the validity state doesn't depend on the model value and (2) that branch is reached by changes in _validation attributes_ rather than the model value, things fall apart.

```vue
<script setup lang="ts">
import { computed, ref } from 'vue';

const val = ref('');
const required = ref(true);
const messageSwitch = ref(false);

const customMessage = computed(() => messageSwitch.value ? 'message B' : 'message A');

function validation(state: ValidityState): string {
  if (state.valueMissing) {
    return 'Enter a value.';
  } else {
    return customMessage.value;
  }
}

const onSubmit = (e: Event) => { alert('Submitted'); e.preventDefault(); };
</script>

<template>
  <form @submit="onSubmit">
    <OField label="Value">
      <OInput v-model="val" :required :customValidity="validation" />
    </OField>
    <OField>
      <OCheckbox v-model="required">Required</OCheckbox>
    </OField>
    <OField>
      <OCheckbox v-model="messageSwitch">Alternate message</OCheckbox>
    </OField>
    <OButton native-type="submit">Submit</OButton>
  </form>
</template>
```

Without `forceValidationReactivityUpdate`, if you blur the input box, you'll get the "Enter a value" message. If you then uncheck "required", you'll get "message A", but toggling the other checkbox doesn't start displaying "message B" like it should.

blm768: I pushed up a fix for another issue I found while testing that example. I also found that the fix allowed for a refactoring that simplifies this logic a bit. The given example would still be broken without forcing the watcher to run again, though.

packages/oruga/src/composables/useInputHandler.ts
blm768: I found an issue that required me to call `checkHtml5Validity` from inside the `watchEffect`, so I had to flip the order of calls around to avoid infinite recursion.

packages/oruga/src/composables/useInputHandler.ts
blm768: I discovered that getting the types to line up was slightly less painful than I thought, although the strength of the actual type checking is pretty limited. I've pushed a fix for this.

mlmoravek:
> Should we consider just adding browser-based testing back in for these kinds of cases? This definitely seems like it's getting complicated enough that it needs browser-level tests.

I've never done browser-based testing before. Feel free to add a new PR with the configuration changes and some primitive tests to demonstrate.

mlmoravek:
Thanks mate, looks good now, great work!

Will you finish it up with all the other components? Or should i do that?

blm768:
> I've never done browser-based testing before. Feel free to add a new PR with the configuration changes and some primitive tests to demonstrate.

Sure; I can do that. Would you like me to do that after this PR, or should I consider the browser-based tests a prerequisite for this change?

> Will you finish it up with all the other components? Or should i do that?

I'll get the rest of the components taken care of. It should be pretty quick to sort them out.

mlmoravek:
> Would you like me to do that after this PR, or should I consider the browser-based tests a prerequisite for this change?

Maybe just finish this one before opening another bigger topic. 

> I'll get the rest of the components taken care of. It should be pretty quick to sort them out.

Gool!

blm768:
Rebased to make the next stage of review easier. I'll extend these changes to all input types next.

blm768:
I've pushed up some new code changes, When it comes to the documentation, I was trying to build the documentation to test my edits and discovered that the Markdown files seem to be automatically generated. Should I make any updates to the documentation, or should I just let that process take care of things?

mlmoravek:
The documentation is generated by `npm run gen` to generate whatever is generated, or just run `npm run gen:docs` to generate the docs, or `npm run dev:watch` to start a local documentation dev server with hmr for the oruga package and one time markdown files generation. 

Please generate the docs and push the documentation changes to this branch as well.

packages/oruga/src/components/autocomplete/Autocomplete.vue
```diff
@@ -226,13 +226,22 @@ const props = defineProps({
...
         default: () => getOption("useHtml5Validation", true),
     },
-    /** The message which is shown when a validation error occurs */
-    validationMessage: { type: String, default: undefined },
+    /** Custom HTML 5 validation error to set on the form control */
+    customValidity: {
+        type: [String, Function] as PropType<
+            | string
+            | ((
+                  currentValue: T | null | undefined,
```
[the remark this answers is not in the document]
blm768: If the `modelValue` prop is missing entirely, it tends to come through as `undefined` instead of `null`.

packages/oruga/src/components/taginput/Taginput.vue
```diff
@@ -536,12 +545,11 @@ defineExpose({ focus: setFocus, value: items });
                 :keep-first="keepFirst"
                 :keep-open="keepOpen"
                 :has-counter="false"
-                :use-html5-validation="useHtml5Validation"
+                :use-html5-validation="false"
```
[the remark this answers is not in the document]
blm768: The parent `Taginput` element uses `useInputHandler`, which drills down into the child component and directly accesses its `<input>` element. With constraint validation enabled for the child, there would be two different instances of `useInputHandler` both trying to operate on the same element, which can cause various issues. For instance, the two calls to `useInputHandler` can "fight" over the custom validation message.

mlmoravek:
It seems that you have an older version of the oruga theme installed. 
Could you try deleting your `node_modules` and perhaps `package-json`, and running `npm install` from the root folder again? 
There shouldn't be any doc changes with theme variables...

blm768:
Thanks for sorting out the documentation! I'll keep that in mind for my next PR.

PR #1133, opened 2024-11-25
fix(performance): solve performance issues with child component registration

### LINKED ISSUE #1057: Performance issue with autocomplete when the data array is very large.
### Overview of the problem

**Oruga** version: [0.9.0-pre.1]
**Vuejs** version: [3.5.4]
**OS/Browser**: windows/chrome

### Description
Performance of Autocomplete is very bad when the data is very large. I have an array which contains around 1000 strings.  It feels very laggy when using the Autocomplete component.

### Steps to reproduce

The base example on the page: https://oruga-ui.com/components/Autocomplete.html.
First fill the data object with random string data. When the data object is large enough, the performance drops significantly.

### Expected behavior

the filter function should perform very smoothly and feel snappy.

mlmoravek:
Fixes #1057

## Proposed Changes

- Extract child item sorting after DOM order into a debounced watch handler
- Remove `sortedItem` and sort `childItems` directly
- Deprecate prop `destroyOnHide` from Tabs component (#1138)
- Deprecate prop `destroyOnHide` from Steps component (#1138)
- Add `active` prop to Tabs / Steps slots


mlmoravek:
@dauriat Could you have a look at my proposal to fix the performance issue?

dauriata:
Thanks for your quick PR, I tested it and I have two observations : 
* with a 3000 array options with autocomplete it is still taking 15s to switch options. I realized that childItems are not even used in the autocomplete components, so we could go a little further and add also a registration option. If skipping the push to childItems array, it takes 2s to switch 3000 entries which seems acceptable to me.

* some childItems are not mounted in DOM, so matching childItems with DOM doesn't work even with a debounce. The debounce actually breaks things even more as the active item, only in DOM,  would always be number 1. Without debounce it was at least in the right order before any insertion / removal of item. Look at the steps example in docs.  
I think the only solution is to either  
  *   find a way to match the list of slots ( I think I got a working solution, I could push you a draft PR, but it is using some internals of vue as was doing buefy) , 
   *  or ask per documentation to provide an ordered index property in the childItem.

dauriata:
You were right I could not use $slots in the parent to retrieve id to match the childs.
Here is a draft commit that I found working with steps ( I didn't look at the other component)
* so the ideas is to ensures the child template is not empty so at least an empty hidden wrapper is mounted in DOM for it to be sorted
* I used a mounted hook instead of a debounce function
https://github.com/dauriata/oruga/commit/7f2d1a55ed9de4c49df3f4d0b0470ed3631082f5

mlmoravek:
> I used a mounted hook instead of a debounce function

But with this you go back and call the `sortItems()` function for each child component. For example, if you have a dropdown with 1000 items, this will be called 1000 times. By debouncing it, it is only called once at the end. I think we need something like debouncing or something similar to reduce the amount of calls to just one after all the childs are mounted, not after each and every one.

> so the ideas is to ensures the child template is not empty so at least an empty hidden wrapper is mounted in DOM 
Ahh ok, thanks for your example, now I understand the problem here :D

dauriata:
> , if you have a dropdown with 1000 items, this will be called 1000 times. 
Oh ok you're right but I went back to idea that with sorting option, sorting will only happen on tabs or steps where it is unlikely to have 1000 items

dauriata:
@mlmoravek 
regarding [e0da019](https://github.com/oruga-ui/oruga/pull/1133/commits/e0da01959340ddd5136bcc5a28562c5d13d020c7) why would you want to remove destroyonHide Prop ? It is a feature that can be useful for lazy loading component. Are you planning to delete it for tabs to ?

PR #1172, opened 2025-01-31
docs: scan for component default scss vars

whiplashwebb:
## Proposed Changes

In order to support https://github.com/oruga-ui/theme-bulma/pull/154 we need to add a secondary mechanism for pulling scss variables from our files, as defaults can no longer be stored with the bulk of the scss for a given component after recent Bulma changes. This change allows the docs to scan `/scss/component-defaults/*` in addition to `/scss/components/*` and merge the results together.

I have tested this change with the current version of theme bulma and with the updated version of theme bulma. Both versions function as expected, so we should be able to merge this PR prior to the theme PR merging. Other themes are unaffected by this change, although they can opt-in to this pattern if needed.

This change only affects documentation and contains no changes to components etc.

### DIFF
--- a/packages/docs/src/templates/style.ts
+++ b/packages/docs/src/templates/style.ts
@@ -3,9 +3,9 @@ import * as fs from "fs";
 import { type SafeDocgenCLIConfig } from "vue-docgen-cli/lib/config";
 import { getThemePath, Themes, type ThemeConfig } from "../themes-helper";
 
-export const docsRegex = "/* @docs */";
+const docsRegex = "/* @docs */";
 
-export function getVariablesFromContent(content: string) {
+function getVariablesFromContent(content: string): string[] {
     const docs = content.substring(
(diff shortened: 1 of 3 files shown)


packages/docs/src/templates/style.ts
```diff
@@ -3,6 +3,20 @@ import * as fs from "fs";
 import { type SafeDocgenCLIConfig } from "vue-docgen-cli/lib/config";
 import { getThemePath, Themes, type ThemeConfig } from "../themes-helper";
 
+export const docsRegex = "/* @docs */";
+
+export function getVariablesFromContent(content: string) {
```
mlmoravek, line 8: Do we need to export this function and de docsRegex?

whiplashwebb: no, that was just force of habit. removed.

PR #1271, opened 2025-04-03
Bug : Input icon no longer shows Bulma hover/focus styles on left icon

whiplashwebb:
Bulma supports adding hover and focus colors to input icons. Oruga has long supported this, but in the recent rewrites this functionality regressed. This PR fixes the problem.

Example of correct behaviour from [Bulma docs](https://bulma.io/documentation/form/input/#with-font-awesome-icons)

https://github.com/user-attachments/assets/1255d326-07c2-4fd1-a38b-4eaeb07f9b4c

Example of incorrect behaviour from Oruga docs

https://github.com/user-attachments/assets/1bb5e9ee-29b8-42f1-9d75-3219e203eba7

## What broke?

Bulma targets this style with a subsequent sibling selector sort of like this

```
.input:focus ~ .icon {
color:red;
}
```

When the input component was re-written the order of the input dom elements changed, going from `input, left icon, right icon` to `left icon, input, right icon`. This is totally logical since DOM order reflects visual order, but it breaks the Bulma selector for the left icon since it now comes **before** the focused input.

Luckily, both Bulma and Bootstrap position icons absolutely, so changing the order doesn't break anything else.

## Proposed Changes

- Changes the order of elements so icons follow input.


mlmoravek:
I don't like adapting the HTML for a theme. The theme should adapt to Oruga's HTML. To achieve this, the theme includes custom CSS in addition to Bulma or Bootstrap.

Maybe the theme could extend the bulma rule with something like: 
```
.icon:has(+ input:focus) {
...
}
```

whiplashwebb:
@mlmoravek I understand your objection and all things being equal I would agree, however I think in this situation i makes more sense to change the markup. 

First off, this markup is consistent with Oruga prior to 0.9.x and Buefy before it. Changing the order of the markup in 0.9.x caused this regression, so I would argue we're fixing a bug rather than altering Oruga to better suit one theme. 

Second, none of our themes care about the order of these elements, as all use absolute positioning for input icons. We don't get any benefit from the current order of DOM elements apart from code aesthetics. I value code aesthetics as much as the next guy, but practical concerns have to come first, and the change in order causes real practical harm here. 

Third, the solution you propose creates a duplicate layer of css which we need to keep in sync with Bulma. This can be a maintenance problem as we just saw in [this theme bulma PR](https://github.com/oruga-ui/theme-bulma/pull/165/files), where our theme styles fell out of sync with the bulma layer and caused a subtle bug. 

Thoughts?

mlmoravek:
I would say, the markup does also matter, but I unserstand your maintainance reason. 

Could you check if Bootstrap itself (not the theme) is also forcing this order (Input, icon, icon) or only Bulma does this ? 

And btw, if the themes don't care about the order of the elements, the order should be any problem ;)

whiplashwebb:
I'm digging through the bootstrap docs right now and I don't think bootstrap itself has this input/icon pattern. There's no icon component in core bootstrap, nor are there any input examples with icons. I forgot how hands-off bootstrap is about icons. They do have form groups, which could be used to create a vaguely similar design, but the markup doesn't overlap. I think this is a Bulma feature which has been added to Oruga's bootstrap implementation for parity?

TLDR : bootstrap looks good.

mlmoravek:
> TLDR : bootstrap looks good.

Then I would say that bulma has some incorrect styling here, and we should prioritise more correct structured markup here.

mlmoravek:
Please change it in the bulma theme or try to fix it in bulma itself.

PR #1279, opened 2025-04-11
fix(notification): solve programmatic does not pass variant to notification

### LINKED ISSUE #1282: Notification variant property is ignored when using programmatic interface
### Overview of the problem

**Oruga** version: [0.10.4]
**Vuejs** version: [3.5.13]
**OS/Browser**: Mac OS 15.3.2/Chrome 135.0.7049.95

### Description

When calling a notification programmatically, as outlined [here](https://oruga-documentation-preview.netlify.app/components/Notification.html), the `variant` property is ignored.

### Steps to reproduce

1. Create a function to programmatically launch a notification, such as:
```js
function showNotification(variant) {
    oruga.notification.open({
        message: "danger notification",
        variant: "danger",
    });
}
```
2. Call the function

### Expected behavior

The notification is rendered using the provided variant.

### Actual behavior

Regardless of what value for `variant` is used, the notification always shows with a default style. Inspecting the notification reveals that the notification has no class attached, i.e. "is-danger":
```html
<article role="alert" aria-atomic="true" data-oruga="notification" class="notification is-bottom"><!----><div class="media"><!----><div class="media-content"><span>danger notification</span></div></div></article>
```

### Additional Info

- I've tried all known variants: primary, info, success, warning, danger, light, dark, etc.
- The notification component `<o-notification>` correctly uses the variant property. 
- Adding `class: "is-danger"` DOES render the component correctly by manually adding the class to the article the way variant is supposed to.

whiplashwebb:
Fixes #1282

If you check the docs, the notification page has this code which should open a success notification.

```
oruga.notification.open({
        message: "Something happened correctly!",
        variant: "success",
        closable: true,
    });
```

This code successfully sets closeable and message but it does not pass variant from the notification notice to the notification, which results in a grey default toast rather than the desired green success toast. Previous versions of Oruga (0.8.x and below) bound the notice's props to the notification by default, which passed along variant. 0.9.x instead binds `$attrs`, which does not contain variant.

To solve this problem I've added a daisy-chain to pass only variant down rather. This is not a particularly elegant solution but I'd rather stick to something conservative rather than risk regression with a more elegant refactor. @mlmoravek let me know if you have a better idea for how to do this.

### DIFF
--- a/packages/oruga/src/components/notification/NotificationNotice.vue
+++ b/packages/oruga/src/components/notification/NotificationNotice.vue
@@ -208,9 +208,9 @@ defineExpose({ close });
         v-model:active="isActive"
         :override="override"
         :position="position"
+        :variant="variant"
         :role="isAlert ? 'alert' : 'status'"
         :aria-atomic="true"
-        :variant="variant"
         @close="close">
         <template #inner="{ close }">
             <!-- injected component for programmatic usage -->
--- a/packages/oruga/src/components/notification/NotificationNotice.vue
+++ b/packages/oruga/src/components/notification/NotificationNotice.vue
@@ -210,6 +210,7 @@ defineExpose({ close });
         :position="position"
         :role="isAlert ? 'alert' : 'status'"
         :aria-atomic="true"
+        :variant="variant"
         @close="close">
         <template #inner="{ close }">
             <!-- injected component for programmatic usage -->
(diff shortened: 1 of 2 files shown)


mlmoravek:
@whiplashwebb Could you add a test for this bug in `useNotificationProgrammatic.test.ts`, so it can never happen again ? :D
 Just add the prop and check if the class get applied in the HTML.

PR #1281, opened 2025-04-19
test: add Playwright browser tests

blm768:
## Proposed Changes

- Tests now include browser tests using Playwright.
- The old Cypress tests are now ported to Playwright.

This establishes a basic browser test framework that we can use to validate complicated UI behavior, e.g. HTML constraint validation.

### DIFF
--- a/.github/workflows/ts-check.yml
+++ b/.github/workflows/ts-check.yml
@@ -14,7 +14,8 @@ jobs:
         with:
           node-version: 22.x
           registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
       - name: Install dependencies 📦
-        run: npm install
+        run: npm ci
       - name: Run typescript check 🧐
         run: npm run test:ts
--- a/packages/oruga/src/components/button/tests/Button.browser.test.ts
+++ b/packages/oruga/src/components/button/tests/Button.browser.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, test } from "vitest";
+
+import { render } from "vitest-browser-vue";
+
+import OButton from "../Button.vue";
+
+describe("<Button>", () => {
+    test("render button text", async () => {
+        const screen = render(OButton, {
+            slots: {
+                default: () => "Hello World!",
+            },
+        });
+        const button = screen.getByRole("button");
+        await expect.element(button).toHaveTextContent("Hello World!");
+        await button.click();
+        const clicks = screen.emitted("click");
+        expect(clicks).toHaveLength(1);
+    });
+});
--- a/packages/oruga/vite.config.ts
+++ b/packages/oruga/vite.config.ts
@@ -32,6 +32,8 @@ const entries = {
     }, {} as any),
 };
 
+const browserTestPattern = "src/**/*.browser.test.{ts,js}";
+
 // https://vitejs.dev/config/
 /** @type {import('vite').UserConfig} */
 export default defineConfig(({ mode }) => ({
@@ -136,5 +138,29 @@ export default defineConfig(({ mode }) => ({
         coverage: {
             provider: "istanbul",
         },
+        workspace: [
+            {
+                extends: true,
+                test: {
+                    exclude: [browserTestPattern],
+                },
+            },
+            {
+                extends: true,
+                test: {
+                    include: [browserTestPattern],
+                    name: "oruga-browser",
+                    browser: {
+                        enabled: true,
+                        provider: "playwright",
+                        instances: [
+                            {
+                                browser: "chromium",
+                            },
+                        ],
+                    },
+                },
+            },
+        ],
     },
 }));
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,8 @@ jobs:
         with:
           node-version: 22.x
           registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
       - name: Install dependencies 📦
-        run: npm install
+        run: npm ci
       - name: Build lib 🛠️
         run: npm run build:lib
--- a/.github/workflows/commit-lint.yml
+++ b/.github/workflows/commit-lint.yml
@@ -16,6 +16,7 @@ jobs:
         with:
           node-version: 22.x
           registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
       - name: Install dependencies 📦
         run: npm install @commitlint/cli
       - name: Run commit message check 💬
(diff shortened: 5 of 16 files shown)


packages/oruga/vitest.workspace.ts
```diff
@@ -0,0 +1,28 @@
+import { defineWorkspace } from "vitest/config";
+
+const browserTestPattern = "src/**/*.browser.test.{ts,js}";
+
+export default defineWorkspace([
```
mlmoravek, line 5: Can't this be integrated in the current existing vite.config file?

.github/workflows/ts-check.yml
```diff
@@ -14,7 +14,8 @@ jobs:
         with:
           node-version: 22.x
           registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
       - name: Install dependencies 📦
-        run: npm install
+        run: npm ci
```
mlmoravek, line 19: What does this change do? Why from install to ci? What does ci do?

blm768: The "`ci`" [is short for "clean install"](https://docs.npmjs.com/cli/v11/commands/npm-ci). It's basically `npm install` but with a couple of key differences:

- It totally wipes out any existing `node_modules` folder (which shouldn't really apply to these jobs).
- If `package-lock.json` is out of date (relative to what's specified in `package.json`), it will raise an error rather than silently updating it.

The NPM documentation recommends using `npm ci` in CI builds.

packages/oruga/vitest.workspace.ts
blm768: As it turns out, it can be integrated.

.github/workflows/ts-check.yml
blm768: Looks like I missed updating one workflow, though.

packages/oruga/src/components/autocomplete/tests/SimpleAutocomplete.vue
```diff
@@ -0,0 +1,35 @@
...
+    "Node.js",
+    "React",
+    "RxJS",
+    "Vue.js",
+];
+const name = ref("");
+const selected = ref<string>();
+</script>
+
+<template>
```
mlmoravek, line 21: Maybe you can convert this component into an inline component, like you can see in `carousel.axe.test.ts`. so we reduce the need for an extra file and have everthing fo the test in one file :)

packages/oruga/src/components/button/tests/Button.browser.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe("<Button>", () => {
+    test("render button text", async () => {
+        const screen = render(OButton, {
+            slots: {
+                default: () => "Hello World!",
+            },
+        });
+        const button = screen.getByRole("button");
+        await expect.element(button).toHaveTextContent("Hello World!");
```
mlmoravek, line 15: why do you need to await the expect here?

packages/oruga/src/components/collapse/tests/SimpleCollapse.vue
```diff
@@ -0,0 +1,47 @@
...
+        text: "Text 2",
+    },
+    {
+        title: "Title 3",
+        text: "Text 3",
+    },
+];
+</script>
+
+<template>
```
mlmoravek, line 24: Maybe you can convert this component into an inline component, like you can see in carousel.axe.test.ts. so we reduce the need for an extra file and have everthing fo the test in one file :)

packages/oruga/src/components/button/tests/Button.browser.test.ts
blm768: `expect.element` is syntactic sugar for [`expect.poll`](https://vitest.dev/api/expect#poll), so it repeatedly (and asynchronously) polls the DOM until the condition is met or it times out. See [the API docs](https://vitest.dev/guide/browser/assertion-api.html) for usage notes.

mlmoravek:
@blm768 Can you explain to me why Oruga needs this test setup? Aren't normal vitest cases not enough in some scenarios?

blm768:
> @blm768 Can you explain to me why Oruga needs this test setup? In which scenarios aren't normal vitest cases not enough?

The current Vitest setup uses [`jsdom`](https://www.npmjs.com/package/jsdom) to simulate the DOM API. It provides many of the basics necessary to simulate element rendering, basic event handling, etc., but there are many browser APIs and behaviors that it doesn't fully simulate. For instance, I don't think it simulates the response of native `<select>` elements to keyboard events, which is an important part of testing `<OAutocomplete>`. The original use case was to [test how Oruga handles HTML constraint validation](https://github.com/oruga-ui/oruga/pull/1023#issuecomment-2252929513), which is complex enough to require a full browser engine. I haven't written the tests for that yet, but if this PR is merged, those tests are next on my list.

blm768:
> @blm768 You have added the test script call to the test workflow.

Are you referring to the change [here](https://github.com/oruga-ui/oruga/pull/1281/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R19-R33)?

> However, how can the playwright test be called locally?

It's automatically included in normal Vitest runs. The changes in the CI builds just install the dependencies for the tests (in particular, the browser).

> Is it better to add the script to the package.json and just call the npm script in the workflow?

We could add a script to `package.json` to install the test dependencies (`npx playwright install chromium`), but the test runner will prompt the user to run that if they haven't done so already. Once the test dependencies are installed, they don't have to be reinstalled for each run, so the test script would only be used infrequently.

mlmoravek:
That's fine :) Finish what you think is open and we can include it.

blm768:
I've squashed that commit in, so this should be ready. Thanks for looking at it!

packages/oruga/vite.config.ts
```diff
@@ -136,5 +138,29 @@ export default defineConfig(({ mode }) => ({
...
             provider: "istanbul",
         },
+        workspace: [
+            {
+                extends: true,
+                test: {
+                    exclude: [browserTestPattern],
+                },
+            },
+            {
```
mlmoravek, line 148: Can you explain to me why we have two separate entries in the workspace here, with the second being an extended variation of the first?

blm768: The two entries allow us to run most of the tests using the previous JSDOM implementation and only start up a full browser for tests that require it (`*.browser.test.{js,ts}`). The entries don't extend each other; they both [extend the root test configuration](https://vitest.dev/guide/workspace.html#configuration), which means we don't have to duplicate the `setupFiles` and `coverage` keys.

blm768: The `setupFiles` and `coverage` seem like they _could_ be applied just to the non-browser configuration, but doing that breaks all the tests for some reason.

PR #1298, opened 2025-05-12
test(input): test input handler

blm768:
## Proposed Changes

- `useInputHandler` now has browser tests for its constraint validation behavior.
- `useInputHandler` scrolls to invalid form controls even if they're outside an `<OField>`.
- Changes to constraint validation attributes correctly trigger validation message updates even when the input didn't previously require a `MutationObserver` to watch for attribute changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1786,14 +1786,11 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.26.7",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.26.7.tgz",
-      "integrity": "sha512-AOPI3D+a8dXnja+iwsUqGRjr1BbZIe771sXdapOtYI531gSqpi92vXivKcq2asu/DFpdl1ceFAKZyRzK2PCVcQ==",
+      "version": "7.27.1",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.27.1.tgz",
+      "integrity": "sha512-1x3D2xEk2fRo3PAhwQwu5UubzgiVWSXTBfWpVd2Mx2AzRqJuDJCsgaDVZ7HB5iGzDW1Hl1sWN2mFyKjmR9uAog==",
       "dev": true,
       "license": "MIT",
-      "dependencies": {
-        "regenerator-runtime": "^0.14.0"
-      },
       "engines": {
         "node": ">=6.9.0"
       }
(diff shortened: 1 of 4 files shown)


packages/oruga/src/composables/useInputHandler.ts
```diff
@@ -351,13 +349,18 @@ export function useInputHandler<T extends ValidatableFormElement>(
...
 
                 // Update the watcher.
                 // Note that this branch is also used for the initial setup of the watcher.
                 // We're assuming that `maybeElement` will start out null when the watcher is created, which will
                 // cause the watcher to be triggered (with `oldEl == undefined`) once the component is mounted.
-                if (needWatcher && isDefined(el) && el !== oldEl) {
+                if (
+                    needWatcher &&
+                    isDefined(el) &&
+                    (validationAttributeObserver == null || el !== oldEl)
```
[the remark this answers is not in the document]
blm768: The behavior I saw in the tests was that:

- The component started with `validationAttributeObserver == null`. `needWatcher` was `false` because `el` was `null`.
- The component was mounted, making `el` non-null, but because the element was valid (or, more specifically, wasn't marked to _display_ a validation error) and wasn't using a custom validator, `needWatcher` was still `false`, leaving `validationAttributeObserver == null`.
- The element was marked as invalid on blur, but the old condition here didn't pass, so we didn't create a watcher, and disabling the input didn't clear the validation message like it was supposed to.

The new condition should correctly handle any case where we've determined that we need a watcher but we either don't have one (`validationAttributeObserver == null`) or aren't watching the right element (`el !== oldEl`).

blm768:
I've rebased this to account for the dependency update on `develop`.

packages/oruga/src/composables/useInputHandler.ts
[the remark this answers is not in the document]
mlmoravek: Ok, but I'm not sure if we need another abortion `if` after the observer initialisation...
What happens here when `el` is null/undefined `validationAttributeObserver.observe(el, ... )` ? 
Also, i think this will raise an error if `el` is empty: 
```
let ancestor: Node | null = el;
while ((ancestor = ancestor.parentNode)) {
   ...
 ```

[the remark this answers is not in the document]
blm768: > Ok, but I'm not sure if we need another abortion `if` after the observer initialisation...

Are you referring to line 363?

> What happens here when `el` is null/undefined `validationAttributeObserver.observe(el, ... )` ?

We'll never call `validationAttributeObserver.observe` if `el` is null or undefined; both the definition of `needWatcher` and [the check on line 361](https://github.com/oruga-ui/oruga/pull/1298/commits/858a288bc2a024f5d2775188c95244f5cc9405e3#diff-e8e2be411f6a4c43f5676dd3bffe86bda2ad8c1420749d6376f37f7b8a75566dR361) will prevent it.

> Also, i think this will raise an error if `el` is empty:

Due to the check on line 361, `el` will never be empty if we reach that loop.

package-lock.json
mlmoravek, line 1: Can you try to remove this package-lock.json changes?

mlmoravek:
For me, the validation observer stuff in the useInputHandler is one of the most complex stuff in the whole project :D

package-lock.json
blm768: I could drop that commit, but autocomplete and in-editor type checking would be broken, at least in the test code. I haven't found another way to fix those. See the message on [the first commit](https://github.com/oruga-ui/oruga/pull/1298/commits/98c9605d24ee2bec434c36868ce5ac7d1605f886) for further details.

blm768:
> For me, the validation observer stuff in the useInputHandler is one of the most complex stuff in the whole project :D

Yep; it's tricky stuff. I found a couple of bugs in it just by writing the tests, so you're not alone in that opinion. Some of the complexity is unavoidable, but we may eventually come up with a clearer way to write the code so it's easier to understand and maintain.

package-lock.json
mlmoravek: Ok. Maybe I got something wrong updating the dependencies.

blm768: I ran into the same issue trying to update the packages myself, so I think it's more that `npm` did something wrong.

PR #1308, opened 2025-05-16
docs(collapse): update documentation

mlmoravek:


mahmost:
It seems theme-bulma is not working here int the deploy preview (this is also the case with currently live docs)

image

mlmoravek:
@mahmost I know. The examples in the documentation are wrapped in a web component. The Bulma theme currently has some issues with setting CSS variables correctly when used in a web component, introduced in https://github.com/oruga-ui/theme-bulma/pull/168.
I think the theme should work normally outside of a web component.

We are looking into it. :)

PR #1324, opened 2025-05-27
Fix tab buttons in forms

### LINKED ISSUE #1323: Tabs inside forms implicitly submit the form
### Overview of the problem

**Oruga** version: 0.10.6 (and probably older versions)
**Vuejs** version: 3.5.13
**OS/Browser**: tested in Firefox on Windows 10; probably applies to all platforms and browsers

### Description

By default, `<OTabs>` will use `<button>` tags for the tabs. However, when a `<button>` with no `type` attribute is inside a `<form>`, it's [implicitly treated as `<button type="submit">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type).

### Steps to reproduce

Using a theme that doesn't override `tabItem.tag` to use something other than `'button'`, construct a component with the following template:

```vue
<template>
  <form @submit.prevent="console.log('submitted')">
    <o-tabs>
      <o-tab-item value="tab1" label="Tab 1">Tab 1</o-tab-item>
      <o-tab-item value="tab2" label="Tab 2">Tab 2</o-tab-item>
    </o-tabs>
  </form>
</template>
```

### Expected behavior

Switching tabs should not submit the form.

### Actual behavior

Switching tabs submits the form.

### Recommended Solution

Ideally, the tabs would either be `<button type="button">` (which won't implicitly submit the form) or `<a href="#' @click.prevent>` (which also avoids the button behavior and is arguably a more descriptive semantic choice since the tabs are for navigation rather than for performing actions with side effects).

blm768:
Fixes #1323

## Proposed Changes

- `OTabs` now generates `<button type="button">` instead of a plain `<button>`.


blm768:
That's a fair point. I've folded in a fix for the other components that used buttons by default (aside from `<OButton>` since it's meant to have the semantics of a normal button).

PR #1605, opened 2026-04-09
feat(dialog): add `returnValue` to component emits

### LINKED ISSUE #1604: Dialog: Allow arguments in emit('close') of template components
### Description
When using `emit('close)` in a component passed to the dialog via the `component` prop when using the programmatic interface, any arguments passed with the close event are discarded.

This prevents the `onClose` prop that can be used in the programmatic interface from being able to distinguish between different types of closes. For example when you have a dialog with three different buttons which each trigger a close and you want to know which one has been pressed to perform the correct action.

This feature is present in the current o-modal where you can pass parameters to the close function and they will bubble up. We want to migrate from the deprecated o-modal to the o-dialog component but not having this feature makes has stopped us for now.

### Code example
In the following example you can see that only the close event itself is logged, not the corresponding action. If you change it to `oruga.modal.open` then the console log will only log the args.
```
<script setup>
import { ref, h } from 'vue'
import { useOruga, OButton } from "@oruga-ui/oruga-next";

const oruga = useOruga();
 
const lastAction = ref(null)
 
function DialogContent(props, {emit}) { 
  return h('div', { class: 'dialog-content' }, [
    h('header', { class: 'dialog-header' }, [
      h('h2', 'Confirm Action'),
      h('p', 'Please choose one of the options below.'),
    ]),
    h('footer', { class: 'dialog-footer' }, [
      h(OButton, { variant: 'light', onClick: () => emit('close','Cancel')}, () => 'Cancel'),
      h(OButton, { variant: 'success', onClick: () => emit('close','Yes')}, () => 'Yes'),
      h(OButton, { variant: 'danger', onClick: () => emit('close','No') }, () => 'No'),
    ]),
  ])
}
 
const openDialog = () => {
  oruga.dialog.open({
    component: DialogContent,
    trapFocus: true,
    onClose: (event,action) => {
      console.log(event,action)
    },
  })
}
</script>
 
<template>
  <div class="dialog-test">
    <o-button variant="primary" @click="openDialog">
      Open Dialog
    </o-button>
  </div>
</template>
```

mlmoravek:
Fixes #1604

## Proposed Changes

- use `<dialog>` [`returnValue`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/returnValue) feature to add an optional string argument to the ODialog component emits

### DIFF
--- a/packages/docs/components/Dialog.md
+++ b/packages/docs/components/Dialog.md
@@ -75,27 +75,27 @@ The component implements the W3C ARIA APG [Dialog (Modal) Pattern](https://www.w
 
 ### Events
 
-| Event name    | Properties                                | Description                                                              |
-| ------------- | ----------------------------------------- | ------------------------------------------------------------------------ |
-| update:active | **value** `boolean` - updated active prop | active prop two-way binding                                              |
-| close         | **event** `Event` - native event          | the event is fired when the dialog has been closed                       |
-| cancel        | **event** `Event` - native event          | the event is fired when the user wish to dismiss the current open dialog |
-| confirm       | **event** `Event` - native event          | the event is fired when the confirm button get clicked                   |
+| Event name    | Properties                                                                                      | Description                                                              |
+| ------------- | ----------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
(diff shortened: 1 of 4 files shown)


DavidVanDeursen (approved):
Looks great! This will fix our issue, thank you very much!

Code looks clean as well 👍.

PR #1616, opened 2026-04-17
fix(table): check all preserves previously checked rows if keep-checked is true

jarana-ryder:
## Proposed Changes
- Fixes check all  resetting `tableCheckedRows` to only include rows from the current page if `keep-checked` is `true`
- Updates table example, adds `Keep Checked` switch to test the prop

https://github.com/user-attachments/assets/9d0b5d79-3fe4-4c64-9a5c-d71b43f4ed1a


mlmoravek:
Linked to #1548

